PR #43, opened 2014-12-19
Add an optional Java 8+ controller handler to use compiled parameter names

gitblit:
Now that we've merged #42, a Java8+ ControllerHandler becomes basically this:

```java
   @Override
    protected String getParameterName(Method method, int i) {

        String name = super.getParameterName(method, i);

        if (StringUtils.isNullOrEmpty(name)) {
            // parameter is not named via annotation
            // try looking for the parameter name in the compiled .class file
            Parameter parameter = method.getParameters()[i];
            if (parameter.isNamePresent()) {
                name = parameter.getName();
            }
        }

        return name;
    }
```

I am struggling to name this module and I'm not sure if you will agree to merge an optional Java8 module anyway.  Suggestions or feedback welcome.


gitblit:
This can wait till Pippo switches to J8.

PR #53, opened 2014-12-30
Allow entities to be automatically instantiated using optional controller annotations

gitblit:
I'm open to suggestions on names for the annotations.

`@Form` -> build entity from request parameters
`@Body` -> body entity from request body
`@Param` -> named request parameter
`@ParamPattern` -> currently used for handling deserialization of dates

Maybe those are good enough but perhaps...
- `@Param` should be `@Named` ?
- `@ParamPattern` should be `@Pattern` ?

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/controller/Body.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/controller/Body.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


decebals:
I know that this feature is important for you and I will try to promote/integrate it. I reviewed javax.ws.rs (jersey) to see if I found something interested related to this PR (annotation names, ...). My opinion is that are many annotation (I speak about javax.ws.rs) to model a very trivial thing. Look at [jersey api](https://jersey.java.net/apidocs/latest/jersey/index.html) and you will see a lot of `*Param` annotations, `DefaultValue` just for a simple conversion.
My feeling is that with all these annotations I cannot resolve all my cases (maybe 80-90% but not all).
So, I am OK with your annotations. Maybe `@Form` is not that exactly (it's like BeanParam from javax.ws.rs) and pattern can be an option  parameter for `@Param`.
I think that `@Named` is not ok because I can have a conflict javax.inject.Named.

Issue #50, opened 2014-12-29
Decide if Pippo apps can be distributed as a WAR


gitblit:
I'm on the fence on this issue.  Supporting inconsistencies with other containers is always a nuisance.  Having a single deployable product that behaves in production exactly as in development is pretty handy.

The biggest thing we lose is container authentication - but I have always implemented my own anyway.  

Either way we need to decide if Pippo apps require their own dedicated container or if they can be distributed as WARs.

decebals:
I made a test to see if I can run pippo-demo as a WAR using an external container (jetty). I can say that everything is OK.

Steps
- uncomment some lines in `pippo-demo/pom.xml`
```xml
<!--
<exclusions>
    <exclusion>
        <groupId>ro.fortsoft.pippo</groupId>
        <artifactId>pippo-jetty</artifactId>
    </exclusion>
 </exclusions>
-->
```
because I don't want pippo-jetty module in my war
```xml
<!--
<plugin>
	<groupId>org.eclipse.jetty</groupId>
	<artifactId>jetty-maven-plugin</artifactId>
	<version>9.2.0.v20140526</version>
	<configuration>
		<scanIntervalSeconds>10</scanIntervalSeconds>
		<stopKey>foo</stopKey>
		<stopPort>9999</stopPort>
		<webApp>
			<contextPath>/</contextPath>
		</webApp>
	</configuration>
</plugin>

<plugin>
	<groupId>org.apache.maven.plugins</groupId>
	<artifactId>maven-war-plugin</artifactId>
	<version>2.5</version>
</plugin>
-->
```
because I want to use `jetty-maven-plugin` to start an external jetty container
- uncomment some lines in `web.xml` from pippo-demo
```xml
<!--
<filter>
	<filter-name>pippo</filter-name>
	<filter-class>ro.fortsoft.pippo.core.PippoFilter</filter-class>
	<init-param>
		<param-name>applicationClassName</param-name>
		<param-value>ro.fortsoft.pippo.demo.SimpleApplication</param-value>
	</init-param>
</filter>

<filter-mapping>
	<filter-name>pippo</filter-name>
	<url-pattern>/*</url-pattern>
</filter-mapping>
-->
``` 
if you want to run other demo application then change `ro.fortsoft.pippo.demo.SimpleApplication` with something else
- run some commands from command line
```
mvn clean install
cd pippo-demo
mvn war:war
mvn jetty:run-war
```
now the application is available at `localhost:8080`

[...]

PR #54, opened 2014-12-30
Expose all WebServerSettings to application.properties and add external files path

gitblit:
This PR exposes all WebServerSettings to the `application.properties` file.  It also slightly revises the serving of external files.  External files are now served on a subpath of the context instead of co-mingling Pippo & external files directly on the context path. (i.e.  serve external files from `/context/ext` not `/context`).  The external files path is configurable via the API or `application.properties`.

Serving from multiple sources on the same exact path can create problems for other `WebServer` implementations.

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/PippoConstants.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/PippoConstants.java
@@ -53,6 +53,8 @@ public final class PippoConstants {
 
     public static final String SETTING_SERVER_CONTEXT_PATH = "server.contextPath";
 
+    public static final String SETTING_SERVER_EXTERNAL_FILES_PATH = "server.externalFilesPath";
+
     public static final String SETTING_SERVER_EXTERNAL_FILES_LOCATION = "server.externalFilesLocation";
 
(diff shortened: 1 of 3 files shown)


gitblit:
The static files support has already been removed per your permission because it has a viable (and better) alternative: ClasspathResourceHandler.

The external static files feature is only available via the Pippo `WebServer` and since WAR distribution (#50)  is a requirement that automatically makes it a special case.  Right?

I don't personally need external static files - however - I have seen requests in discussion groups for other frameworks on how to serve external files.  I think it makes sense to keep it around for those who want to use it.

Having said that, right now `pippo-jetty` serves external files on the root of the container.  There are two problems with this.

1. the contextPath setting is ignored which will break url references
2. not all potential `WebServer` implementations can gracefully accommodate that scenario

I'd like to specify the path which external files are served from and this path should be relative to the context path.  i.e. pippo serves on `/context/` and external files are served on `/context/ext`.  Having a discrete path allows attaching a handler.

Alternatively, we can drop the external files altogether.  It's currently only useful for serving completely public files.

decebals:
> 1. the contextPath setting is ignored which will break url references

It's a bug

Sure we can preserve this functionality.

PR #56, opened 2015-01-07
Relax the validation of HTTP methods on route registration

gitblit:
Despite the availability of API in Router to add a handler for any HttpMethod, `DefaultRouter` rejects routes if they do not specify one of HEAD, GET, PATCH, POST, PUT, or DELETE.

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/route/DefaultRouter.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/route/DefaultRouter.java
@@ -15,7 +15,7 @@
  */
 package ro.fortsoft.pippo.core.route;
 
-import ro.fortsoft.pippo.core.HttpConstants;
+import ro.fortsoft.pippo.core.PippoRuntimeException;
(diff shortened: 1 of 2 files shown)


decebals:
Is it a reason for this relaxation?

gitblit:
Two reasons.

1. We prevent implementation of HTTP-based protocols (e.g. CalDAV)
2. I want to abuse route matching by using custom methods.  The immediate motivation is to support better filtering.  Your current concept of a filter requires me to specify every HTTP method I want to apply the filter too.  That's a little painful for protecting an API.  If we allow registering custom methods then I can workaround the lack of a real filter.  Spark works this way internally i.e. `addRoute(uriPattern, "BEFORE", handler);`

decebals:
Are many things that are related to the filter concept (for example #60, #62). We will try to find the best approach/implementation for this concept. I want to see if this relaxation brings some improvement.

gitblit:
This definitely helps.

decebals:
> Your current concept of a filter requires me to specify every HTTP method I want to apply the filter too.

Sure. I totally agree with you.

PR #57, opened 2015-01-07
Add RedirectHandler and TemplateHandler to reduce boilerplate

gitblit:


gitblit:
> I see that in this situation (using a TemplateHandler for /contacts ) we cannot use @Metered.

Ah.  Tricky.  I hadn't considered that.  I have some ideas on how to address it but that is a different & larger change.

PR #58, opened 2015-01-07
Fix bug in chain handling where the Response is committed half-way through

gitblit:


decebals:
I don't know here. Maybe the next handler is a filter that has not the intention to commit the response (for example a filter that close the connection to database). Does that make sense?

gitblit:
That makes good sense.

Spark & Ninja don't have this issue because they don't immediately reply to the client from within their Response wrappers (`Request.commit()`).  On those frameworks, if the handler is called, it can modify the `Response`.

Perhaps we should consider pushing the `commit()` capability out of `Response` and into `PippoFilter`.  Additionally, we could implement what Spark has - `PippoHaltException` which is thrown by `response.halt(statusCode)`.

decebals:
Yes I know about `Result` from Ninja and who Spark resolves a request. My intention for pippo is to be a self explained framework with few easy to learn concepts). For example my intention was to avoid adding a `Filter` concept and to add possibility to use a RouteHandler as a filter if I want. For this reason I added  RouteHandlerChain. I think that right now we can interrupt  the chain with an exception in `RouteHandler.handle()`. It's the current route handler decision if it wants to break the chain or not and this is particular for each application. For example if I know that the last route handler in my application is a  filter that close the connection to a database then I don't want to break the route handler chain. Sure, accidentally, in the process of development I can receive an error related to the response already committed  but I know that it's my fall as developer (many route handlers mapped on the same uriPattern, handlers that attend to commit the response).
Sure it's only my theory and maybe it's wrong in practice.

gitblit:
Ok.  I have come around to your line of thinking that breaking the chain - as implemented in this PR by checking isCommitted() - is a bad idea.

Throwing an exception is exactly what I did in #62 but that results in sending an `Error` instance instead of a `Result`.  I'll think more on that and revise #62.

PR #59, opened 2015-01-08
Add default XML content type engine based on JAXB

gitblit:


decebals:
Nice. I think that it's better to move `JaxbEngine` in core (as `TextPlainEngine`).

gitblit:
Moved.

PR #60, opened 2015-01-08
Improve supported Response content types with builder-style setters

gitblit:
Also introduce a `negotiate()` method to streamline handlers:

```java
response.xml().negotiate(request).send(myObject);
```

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
@@ -388,12 +388,62 @@ public class Response {
         return httpServletResponse.isCommitted();
     }
 
+    /**
+     * Attempts to negotiate the Response content type based on the Request
+     * accept type. If an agreement can't be reached, the Response content-type
+     * is not set.
+     *
+     * This allows a response chain such as
+     * <code>response.xml().negotiate(request).send(myObject);</code> which will
+     * set the content type as <code>application/xml</code> and optionally
+     * override it after negotation.
+     *
+     * @param request
+     * @return the response
+     */
+    public Response negotiate(Request request) {
+        ContentTypeEngine engine = contentTypeEngines.getContentTypeEngine(request.getAcceptType());
+        if (engine == null) {
+            log.warn("Failed to negotiate a content type for accept-type '{}'", request.getAcceptType());
+            return this;
+        }
+        return contentType(engine.getContentType());
+    }
+
+    /**
+     * Sets the Response content-type to text/plain.
+     */
+    public Response text() {
+        return contentType(HttpConstants.ContentType.TEXT_PLAIN);
+    }
+
     public void text(Object object) {
         send(object, HttpConstants.ContentType.TEXT_PLAIN);
     }
 
+    /**
+     * Sets the Response content-type to text/html.
+     */
+    public Response html() {
+        return contentType(HttpConstants.ContentType.TEXT_HTML);
+    }
+
+    /**
+     * Sets the Response content-type to application/json.
+     */
+    public Response json() {
+        return contentType(HttpConstants.ContentType.APPLICATION_JSON);
+    }
+
     public void json(Object object) {
         send(object, HttpConstants.ContentType.APPLICATION_JSON);
+    }
+
+    /**
+     * Sets the Response content-type to application/xml.
+     */
+    public Response xml() {
+        return contentType(HttpConstants.ContentType.APPLICATION_XML);
     }
 
     public void xml(Object object) {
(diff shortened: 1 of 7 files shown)


pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
```diff
@@ -388,14 +388,64 @@ public boolean isCommitted() {
...
+     *
+     * This allows a response chain such as
+     * <code>response.xml().negotiate(request).send(myObject);</code> which will
+     * set the content type as <code>application/xml</code> and optionally
+     * override it after negotation.
+     *
+     * @param request
+     * @return the response
+     */
+    public Response negotiate(Request request) {
```
decebals: I had difficulty with this method. Perhaps because the name is not so suggestive. Maybe `contentType(Request request)` is better.

pippo-demo/src/main/java/ro/fortsoft/pippo/demo/SimpleApplication.java
```diff
@@ -40,7 +40,7 @@ public void init() {
 
             @Override
             public void handle(Request request, Response response, RouteHandlerChain chain) {
-                response.send("Hello World");
+                response.text().send("Hello World");
```
decebals: Maybe it's to much and redundant. I like the short version `response.send(myText)`. I think that we must promote always the simple version.

pippo-demo/src/main/java/ro/fortsoft/pippo/demo/SimpleApplication.java
```diff
@@ -122,7 +138,7 @@ public void handle(Request request, Response response, RouteHandlerChain chain)
 
                 Map<String, Object> model = new HashMap<>();
                 model.put("greeting", message);
-                response.render("hello", model);
+                response.html().render("hello", model);
```
decebals: The same comment. It's more direct `response.render()` but as long as both are OK who cares :)

pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
decebals: I have an idea to set the content type in constructor using the request or the accept type (String) - add a new parameter in constructor - but I don't know if it's valuable.

decebals: To be honest I don't understand the scenario (I have the feeling that it's related to REST API) for a construction like `xml().contentType(request)`. It's not so predictable the final value of contentType.

pippo-demo/src/main/java/ro/fortsoft/pippo/demo/SimpleApplication.java
gitblit: As you say, it's not necessary because we default to it with `render()`.  I was just being specific.  When mixing API handlers and page handlers in the same class being explicit about return types may make errors more obvious.

pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
gitblit: Yes, it's exactly for REST API and after thinking about is some more I may have to adjust my implementation and revise the fix I implemented in a97cb2a.

Let's say I want to serve objects for both reads and writes to an API.  I have a few couple choices here.

1. response.send(contact, fixedContentType).
2. response.send(contact, request.getAcceptType())
3. response.send(result, request.getContentType())

The first one is easy, Pippo dictates content type for all responses.  That's fine if the API serves only one specific type (i.e. JSON).

In the second one, Pippo sends the object using the type preferred by the `Request` based on the `Accept` header.  This is the GET/OPTIONS/HEAD scenario.  If the `Accept` header is malformed, empty, or does not report a known content type, we throw a server-side exception which results in an internal error (500) being sent to the client.

In the third one, Pippo sends the object using the type preferred by the `Request` based on the `Content-Type` header of the request.  This is the PUT/POST/PATCH scenario.

The `negotiate` method should be used to gracefully figure out which type to use when sending the object back to the client.  This method needs to consider the HTTP method, first, to determine the header check order.  If I PUT as JSON, then I should receive JSON back.  That's the mistake I made in a97cb2a.  Scenario 2 was failing so I reversed the code but in doing that I think I broke scenario 3.

If you search for *content negotiation* with your favorite search engine you find lots of resources.

So, yes the final value of contentType is not 100% predictable.  But it is guaranteed to be either your specified default (`xml()`) or one of your registered content type engines via `negotiate()`.

gitblit:
Revised.

pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
decebals: Thank you for detailed and clear explanation.

PR #62, opened 2015-01-08
RFC: Add Response.halt() to terminate the RouteHandlerChain

gitblit:
Here is a `halt()` implementation that lets the `ErrorHandler` deal with the situation.  That's probably not right but it highlights what I think is a problem with the `Response` mechanism.  All the `send()` methods are suspect too - anything that commits from `Response`.

If you are writing API handlers you likely want them to return a `Result` object with a status code and an optional message.  Not all of those results are errors, but some are.  For example maybe I want to return:

```java
response.xml().negotiate(request).sendOk("Saved {}", contact.getId());
```

BUT still close my open DB connection in the filter that should execute afterwards.  I recognize that will work today, but are we allowing people to break stuff by committing within the chain and not warning/failing afterwards.  It would be painful to call `checkCommitted()` for every possible setter.  In the filter that closes the DB I can set a header which will not be sent to the client because the response was committed in `sendOk`.

I'm open to suggestions on improving this.  I think we should consider following Spark's request processing loop.  I also think we need to consider a `Result` class which is returned for unspecified bodies or something like that.

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/DefaultErrorHandler.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/DefaultErrorHandler.java
@@ -117,7 +117,14 @@ public class DefaultErrorHandler implements ErrorHandler {
             response.bind("stacktrace", stackTrace);
         }
 
-        handle(HttpConstants.StatusCode.INTERNAL_ERROR, request, response);
+        // assume an internal error unless this is a HaltResponseException
+        int statusCode = HttpConstants.StatusCode.INTERNAL_ERROR;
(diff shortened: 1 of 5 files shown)


gitblit:
After thinking about it some more, I don't think we need *this* PR.  Either we send a Result and break the chain or we throw an exception and send an Error.  I'm going to explore revising the `send` methods.  

Perhaps the best solution would be to allow consumers of Pippo to optionally define a subclass of `Response`.  This way the implementation could be extended without adding a ton of methods to the base Pippo `Response` class.  Of course if you consider doing that then it would make sense to allow definition of a subclass of `Request`.

PR #64, opened 2015-01-11
RFC: Add factories for Request and Response

gitblit:
This allows consumers of Pippo to implement higher level RouteHandlers which contain domain-specific logic such as object serialization/deserialization, etc.

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/DefaultRequestFactory.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/DefaultRequestFactory.java
@@ -23,7 +23,7 @@ import javax.servlet.http.HttpServletRequest;
  * @author James Moger
  *
  */
-public class DefaultRequestFactory implements RequestFactory {
+public class DefaultRequestFactory implements RequestFactory<Request> {
 
     @Override
     public Request createRequest(HttpServletRequest httpServletRequest, Application application) {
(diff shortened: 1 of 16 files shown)


decebals:
I like the idea. 
I do not know if somehow not abuse `Initilizer` and `ServiceLocator`.
Another idea is to move RequestFactory and ResponseFactory in Application (with getter and setter).  I see the Application class as an entry point for my web application. I can adjust many aspects of my application in this class (or overwrite this class). In this way we can keep the class PippoFilter small.
No problem, I agree with the current implementation I just to point out some of my thoughts.

gitblit:
1. The factories must exist in `PippoFilter`, not `Application` and we have no access to `PippoFilter` from anywhere so use of `ServiceLocator` seems like a good idea.
2. A factory may need to be configured from `PippoSettings` or some other source.  Initializing the factories with `Initializer` and `Application` gives that configuration context.  I don't have a specific use-case for that capability at the moment, but that seems like a forseeable need.

gitblit:
I just pushed an improvement to use generics which simplifies the implementation of handlers for the custom Request/Response classes.

decebals:
> The factories must exist in PippoFilter, not Application
Why? In PippoFilter I have access to Application and I can call `application.getRequestFactory().createRequest()`. In this mode in `Application.init()` if I need I can set the `RequestFactory` with setRequestFactory(new CustomRequest()). It's more readable for a developer and it's no black magic.

gitblit:
Ok.  Then we should move all factories out of `PippoFilter` and into `Application`.

Issue #66, opened 2015-01-12
Rethink Response.commit()


gitblit:
In `Response.commit()` we are flushing the buffer which forces *chunked transfer encoding* and finalizes the servlet response.

It feels wrong that we are doing this.  I can't point to documentation that says this is a problem, but I can look at other frameworks and see they do not flush anything and thus do not force chunked encoding.

I think we need to consider some alternative (internal) mechanism for finalizing a response.

gitblit:
Hmm.  Maybe this is not a problem.
http://zoompf.com/blog/2012/05/too-chunky

I am still unsure about committing mid-chain.

decebals:
Initial `Response.commit()` was called in `send`, `file`, `resource`. Means that the response is sent and I cannot do something useful in the next route handlers from chain (if in my route handler that committed the response I call `chain.next` to the end of handle method).
I want to talk more about filters.
Now a `Filter` is a RouteHandler from a `RouteHandlerChain`.  We can split the filters in two types: 
- before
- after 

A `before` filter can set some parameters on response (I don't know if it's ok to commit the response in this stage). An example of the before filter is a possible `AuthenticationFilter`.
Another example of the before filter is a possible `BeginTransactionFilter` (open a connection to a database and start a transaction on each request).
An `after` filter also can set something  on Response but it's a non sens because the response was committed. 
Another example of the before filter is a possible `EndTransactionFilter` (commit/rollback the transaction started in BeginTransactionFilter).
Anytime in a RouteHandler I know if makes sense to call `chain.next` (to call the next handler in chain) or to break the chain (simple don't call chain.next).
I know that some (java) web frameworks with the same profile as pippo (SparkJava, Ninja) use a wrapper object (Object in spark and Result in ninja) and at the end of request handling, in the filter (in case of spark because ninja it's not servlet base), the framework try to send the content/body converting that wrapper object in string. I prefer to expose the Response class to the route handler and allow that to commit/complete the response because using this approach I can send data from an InputStream,  File, things that are not possible in Spark, Ninja. I prefer to be closer to servlet, to expose the request response nature.
Sure maybe my idea is not sustainable to the end and in this context we must change these concepts. 
For now in my tests I see one big issue with the actual concept: I cannot register a filter for many/all http methods in a single route.
I would love to hear what is your opinion, if you have some issues related to the actual request response cycle (route handler), if you consider that we can improve this concept and to make it more clear. 
[...]

PR #69, opened 2015-01-14
Separate the status code convenience methods from the commit action

gitblit:
The status code methods are now setters which follow the builder pattern used by the header, cookie, and content type methods.

Also ensure we are not attempting to modify the response after it has been committed.

This is related to issue #66.

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
@@ -62,12 +62,14 @@ public class Response {
     }
 
     public Response contentType(String contentType) {
+        checkCommitted();
         httpServletResponse.setContentType(contentType);
 
         return this;
     }
 
     public Response characterEncoding(String charset) {
+        checkCommitted();
         getHttpServletResponse().setCharacterEncoding(charset);
 
         return this;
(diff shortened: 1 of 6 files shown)


pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
```diff
@@ -62,12 +62,14 @@ public String getContentType() {
     }
 
     public Response contentType(String contentType) {
+        checkCommitted();
```
decebals: I want to clarify one point. This is not necessary in terms of the servlet theory and it is only a protection for developers (means that theoretically you can change the content type after commit but it's a non sense). Is it right what I say?

pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
```diff
@@ -62,12 +62,14 @@ public String getContentType() {
...
         httpServletResponse.setContentType(contentType);
 
         return this;
     }
 
     public Response characterEncoding(String charset) {
+        checkCommitted();
```
decebals: The same situation

pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
gitblit: Yes.  But I'd like to make it clear that attempting to modify the response after commit() is a no-no.  I think that is better than thinking you can, failing silently, and then debugging/scratching your head.  Of course that is an opinion, not truth.

decebals: Sure. Great.

Issue #70, opened 2015-01-14
Running CrudNgApplication as WAR


decebals:
I try to run CrudNgApplication from pippo-demo as WAR on a tomcat 7. The problem is that I encountered a strange issue:
- sometime (in fact in most cases) in the console browser I see:

> Resource interpreted as Stylesheet but transferred with MIME type application/octet-stream: "http://localhost:8080/webjars/bootstrap/3.3.1/css/bootstrap.min.css".
login:41 Resource interpreted as Script but transferred with MIME type application/octet-stream: "http://localhost:8080/webjars/jquery/1.11.1/jquery.min.js".
login:42 Resource interpreted as Script but transferred with MIME type application/octet-stream: "http://localhost:8080/webjars/bootstrap/3.3.1/js/bootstrap.min.js".
login:43 Resource interpreted as Script but transferred with MIME type application/octet-stream: "http://localhost:8080/webjars/angularjs/1.3.6/angular.min.js".

and of course that I cannot use the webjars defined in head
- sometime I see an error related to transfer encoding chunked (I will try to catch and display the error)

decebals:
The second error is:
> ERR_INVALID_CHUNKED_ENCODING

This error appears only in Chrome.

balamaci:
I see the mimetype is set correctly at the point of ClasspathResourceHandler:123 however 
however I think it's being rewritten in
Response: public void resource(InputStream input) {
           if (isHeaderEmpty(HttpConstants.Header.CONTENT_TYPE)) {
            header(HttpConstants.Header.CONTENT_TYPE, HttpConstants.ContentType.APPLICATION_OCTET_STREAM);
        }
}
because the condition isHeaderEmpty might evaluate to true on Tomcat.

decebals:
tested on tomcat 7.0.57

PR #72, opened 2015-01-14
Fix https://github.com/decebals/pippo/issues/70 to check if not contentt...

balamaci:
...ype was not already set on the response before we overwritte the header instead of letting the container handle it.

### DIFF
--- a/pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
+++ b/pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
@@ -576,7 +576,7 @@ public class Response {
     public void resource(InputStream input) {
         checkCommitted();
 
-        if (isHeaderEmpty(HttpConstants.Header.CONTENT_TYPE)) {
+        if (isHeaderEmpty(HttpConstants.Header.CONTENT_TYPE) && getHttpServletResponse().getContentType() == null) {
             header(HttpConstants.Header.CONTENT_TYPE, HttpConstants.ContentType.APPLICATION_OCTET_STREAM);
         }
 
@@ -670,7 +670,7 @@ public class Response {
 
     private boolean isHeaderEmpty(String name) {
         String value = getHttpServletResponse().getHeader(name);
-        return (value == null) || value.isEmpty();
+        return StringUtils.isNullOrEmpty(value);
     }
 
     private void checkCommitted() {


pippo-core/src/main/java/ro/fortsoft/pippo/core/Response.java
```diff
@@ -576,7 +576,7 @@ public void send(CharSequence content) {
     public void resource(InputStream input) {
         checkCommitted();
 
-        if (isHeaderEmpty(HttpConstants.Header.CONTENT_TYPE)) {
+        if (isHeaderEmpty(HttpConstants.Header.CONTENT_TYPE) && getHttpServletResponse().getContentType() == null) {
             header(HttpConstants.Header.CONTENT_TYPE, HttpConstants.ContentType.APPLICATION_OCTET_STREAM);
         }
```
gitblit: We should not even look at the header for content type.
```java
if (getContentType() == null) {
    contentType(HttpConstants.ContentType.APPLICATION_OCTET_STREAM);
}
```

should also change the `commit()` method to use the same approach.

decebals: The documentation says:
>     /**
     * Gets the value of the response header with the given name.
     * 
     * <p>If a response header with the given name exists and contains
     * multiple values, the value that was added first will be returned.
     *
     * <p>This method considers only response headers set or added via
     * {@link #setHeader}, {@link #addHeader}, {@link #setDateHeader},
     * {@link #addDateHeader}, {@link #setIntHeader}, or
     * {@link #addIntHeader}, respectively.
     *
     * @param name the name of the response header whose value to return
     *
     * @return the value of the response header with the given name,
     * or <tt>null</tt> if no header with the given name has been set
     * on this response
     *
     * @since Servlet 3.0
     */
    public String getHeader(String name); 

and
>     /**
     * Returns the content type used for the MIME body
     * sent in this response. The content type proper must
     * have been specified using {@link #setContentType}
     * before the response is committed. If no content type
     * has been specified, this method returns null.
     * If a content type has been specified, and a
     * character encoding has been explicitly or implicitly
     * specified as described in {@link #getCharacterEncoding}
     * or {@link #getWriter} has been called,
     * the charset parameter is included in the string returned.
     * If no character encoding has been specified, the
     * charset parameter is omitted.
     *
     * @return		a <code>String</code> specifying the
     *			content type, for example,
     *			<code>text/html; charset=UTF-8</code>,
     *			or null
     *
     * @since Servlet 2.4
     */  
    public String getContentType();

So we can use our `Response.getContentType() in all our checks - internally this method calls `httpServletResponse.getContentType()`

We should use the same strategy for contentLenght

gitblit: `Content-Length` is in a funny situation right now.  This is directly related to flushing in `commit()`.

`Content-Length` is an HTTP 1.0 header and it is based on sending the entire response in one shot.  When you flush the response in `commit()` that changes things and sets transfer encoding to *chunked* which is a streaming mode for HTTP 1.1.  This lets the container buffer the response and send pieces over the connection.  When the container is chunking, I believe it removes the `Content-Length` header from the response.  So at the moment I don't think there is any value to this method.

decebals:
Thanks Serban and James for your support

PR #77, opened 2015-01-25
add StaticResourceHandle

decebals:
Now are two implementations available `ClasspathResourceHandler` (serves classpath resources) and `FileResourceHandler` (serves file resources)
I improved pippo-demo-basic with a new route:
```java
// send files from a local folder (try a request like 'src/main/java/ro/pippo/demo/basic/BasicApplication.java')
GET(new FileResourceHandler("/src", "src"));
```
What you think?


gitblit:
I like it.  I made a few tweaks to fix some errors.  Using this we can drop the external files feature?

decebals:
@gitblit Thanks

decebals:
> Using this we can drop the external files feature?

Sure. If you consider that has value we can keep it.

gitblit:
Nope, I'd rather kill it in favor of this approach.

PR #80, opened 2015-01-26
Improvements & fixes to the FileResourceHandler

gitblit:
The file resource handler could not serve anything outside `user.dir` and it should accept a File argument.


decebals:
Thanks

PR #87, opened 2015-02-05
Allow to use a custom session

decebals:
`Session` is a an interface with a `DefaultSession` implementation based on `HttpSession`. You can create an use in your application a custom session (for example a signed cookie session).


gitblit:
Great.  This flexibility is definitely on my want list.  I pushed a change to simplify the `Session` interface.

decebals:
Thanks

gitblit:
I just pushed an important fix to this branch after you merged.

The problem is, Request.getSession(false) will create a DefaultSession with
a null http backing session.  The SessionFactory must create a complete
session otherwise you end up with NPEs downstream.

PR #88, opened 2015-02-05
SessionFactory must create sessions

gitblit:
SessionFactory must create a complete session.  If "create" is passed as a parameter
then DefaultSessions can be created with a null httpSession which is wrong.

The correct behavior is for the Request to control creation of the Session.

Refs #87.


decebals:
I think that we have a problem in this scenario:
- first request from browser => a RouteHandler calls Request.getSession()
- second request from browser => a RouteHandler calls Request.getSession(false)
In the second request I retrieve a null session.

gitblit:
Hmm.  Ok.  You are right.

PR #90, opened 2015-02-09
Flash

decebals:
Please tell me if you see some issues before to merge.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/DefaultSession.java
+++ b/pippo-core/src/main/java/ro/pippo/core/DefaultSession.java
@@ -61,6 +61,16 @@ public class DefaultSession implements Session {
         httpSession.invalidate();
     }
 
+    @Override
+    public Flash getFlash() {
+        Flash flash = get("flash");
+        if (flash == null) {
+            put("flash", flash = new Flash());
+        }
+
+        return flash;
+    }
+
     public boolean isNew() {
         return httpSession.isNew();
     }
--- a/pippo-core/src/main/java/ro/pippo/core/Flash.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Flash.java
@@ -15,6 +15,8 @@
  */
 package ro.pippo.core;
 
+import ro.pippo.core.util.StringUtils;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -40,10 +42,6 @@ public class Flash implements Iterable<Flash.Message> {
         messages.add(new Message(level, message));
     }
 
-    public void error(String message) {
-        add(Message.ERROR, message);
-    }
-
     public List<String> get(int level) {
         if (messages.isEmpty()) {
             return Collections.emptyList();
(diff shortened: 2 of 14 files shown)


gitblit:
1. `Flash` & `Message` should be `Serializable`.
2. You have a `getFlash()` method on `Session`.  Should the put("flash") design be an implementation detail of `DefaultSession`?  That would allow a cookie-based session to handle Flash as a cookie too.

decebals:
> Should the put("flash") design be an implementation detail of DefaultSession?

I don't understand.

decebals:
@gitblit I wish to merge this PR but I am waiting a response from you (see my previous comment).

gitblit:
You created a method Session.getFlash() but in PippoFilter.processFlash()
you access it as session.get("flash") or session.remove("flash").

I think we should either drop the getFlash() method and stick with named
keys OR we should add api to let Session implementations manage how Flash
is stored/retrieved/removed.  I don't have a strong preference one way or
the other - but I think it should be consistent.

decebals:
> You created a method Session.getFlash() but in PippoFilter.processFlash()
you access it as session.get("flash") or session.remove("flash").

We agreed that `flash` it's an attribute of session (`session.getFlash()` is equivalent with `get("flash")`). In `PippoFilter.processFlash()` (that is an internal method) I want to retrieve the actual flash from the current session (if it exists), to put this flash in Response.locals` with `response.bind("flash", flash)` and to prepare the outcoming flash with `session.put("flash", new Flash())`.
I used your nice implementation for `session.remove("flash"):T` to make a clone and clean in the same step. Sure I can add a `clean` (or maybe a better name) method in Flash and another constructor that takes a Flash (or a messages list) as parameters but I didn't want to add more methods (in signature of the class) that are not valuable for a user (for example a method as clean would be used only internally by PippoFilter).

> I think we should either drop the getFlash() method and stick with named
keys OR we should add api to let Session implementations manage how Flash
is stored/retrieved/removed.  I don't have a strong preference one way or
the other - but I think it should be consistent.

I think that the official method (what we will write in documentation) to take the flash should be  `session.getFlash()` but also `session.get("flash")` it's another possibility and can be used internally.
Related storage I think that flash it's a simple attribute so I don't see a problem. Here we talk about session's storage. Sure we have DefaultSession but should be nice to add a 'stateless' session - for example (Signed)CokieSession.

Am I wrong?

decebals:
Asap I will try to implement a session based on cookie.

gitblit:
The Cookie session is what I am concerned about.

DefaultSession.  I agree "flash" key stored as a session attribute.

CookieSession.  I'm hesitant.  I hate to keep saying Ninja, Ninja, Ninja, but.... they store Flash as a cookie separately from the Session cookie.  Cookies have finite sizes and it may make sense to implement a flash cookie separately from the session cookie.  I'm not sure.  I guess a Cookie session could intercept `put("flash")` and make it behave differently.  That is my hangup.  By using `put("flash")` and `remove("flash")` you are strongly suggesting an implementation which may not map to all potential implementations of `Session`.

We can always change it later if there is a shortcoming so go ahead and merge.  What you have looks good.

decebals:
Thanks James for your valuable comments/feedback.
I am in the middle of a `CookieSession` implementation (RFC). Maybe tomorrow I will have something that works. We will see if we must improve the current flash implementation.
By the way I am not against Ninja. Definitely it is a great library. I am happy to hear from you how other people solve similar aspects.

Issue #92, opened 2015-02-13
CookieSesion implementation


decebals:
I created a new branch `cookie_session` where I try to create a proof of concept for a session implementation based on cookie.
I will try to put in this issue all problems/inconvenient found by me.
For now I found this problem:
- in `CookieSession` I need to have access to response (to save the session cookie in response)

I see these possible solutions:
1. Solution with `RouteContext` (#91) but in the form:
```
public interface RouteContext {
    getRequest(); // or request()
    getResponse(); // or response()
    next()
}
```
The idea is to keep `Request`, `Response` concepts and to add more helper merthods (for example getApplication() or getPippoSettings()). Also I think that it's good to move `getSession` method from request and to put this method in `RouteContext` for example. In this mode I can change the signature of `SessionFactory.getSession(Request, boolean)` to `getSession(RouteContext)`.

I wrote in a previous comment:
> This RouteContext object maybe is valuable in the future if we want to put some business (methods) - of course is not ok to add more parameters in handleRoute and in this scenario we must create this context object (bucket).

Maybe it's that time :smile: 

2. Put on thread local (see `Application.get()`) other information like request (Request.get()) , response, session

First solution is the Ninja way, it's simple to implement but RouteContext has to walk through the entire application as parameter. The second solution is the Wicket way, it's simple to implement, you can access these objects from anywhere in application but probably will be sensible in async mode (if you plan to do something).

What is your opinion? Do you see other options?

gitblit:
Generally, I favor APIs over ThreadLocals.

ThreadLocals have their place and it's great the JVM gives us that option but they are a crutch.  `RouteContext` is a massive change, but I think long-term we'll be happier adopting `RouteContext`.

Like I wrote in #91, I'm in favor of staying close to the servlet api and I would not be opposed to dropping the custom `Request` and `Response` classes along with their factories in favor of native `HttpServletRequest` and `HttpServletResponse`.

Having a `RouteContext` interface and a `DefaultContext` implementation with a factory would be a good idea.  This approach also makes the *generics* idea mostly moot.

gitblit:
This PR for Ninja is interesting and relevant:
https://github.com/ninjaframework/ninja/pull/313

decebals:
Nice.
[...]

Issue #91, opened 2015-02-09
RFC: Alternate signature for RouteHandler


gitblit:
How do you feel about adopting one of the following signatures:

```java
void handleRoute(Context context, RouteHandlerChain chain);
```

or my preference

```java
void handleRoute(Context context);
```

where `Context` is an interface like:
```java
interface Context {
    Request request();
    Response response();
}
```

If we put the chain inside the `Context` then `Context` would need a `next()` method.

```java
void handleRoute(Context context) {
    if (someCondition) {
        context.next();
    }
}
```

I would drop the `Request` and `Response` factories and implement a `ContextFactory`.

I would also move most/all functionality out of `Request` and `Response` into `Context` and not wrap HttpServletRequest/Response at all.

```java
interface Context {
    HttpServletRequest request();
    HttpServletResponse response();
    void next();
    // refactored methods from Request
    // refactored methods from Response
}
```

We have alot of boilerplate with passing around three args all the time.  Customization often requires having both `Request` and `Response` so I would prefer to merge all these concepts into one type/argument.  From what I see most other frameworks have also come to this conclusion: Wicket, Ninja, Ratpack, etc.

I can live with two args if you are opposed to putting the chain inside the `Context`.

decebals:
Personally I prefer to materialize both concepts `Request` and `Response` and to not hide the Request-Response nature. 
I don't want to be another Play, Ninja implementation and to clone these frameworks. 
Pippo is something different, you can read some parameters from the request and to send the response directly. In ninja for example, I feel that Context it's like a __DTO__, a big bucket, a storage for all response parameters, helper methods for both request and response. You cannot send a response directly (the framework creates a response from a Context and send this response, you cannot access the response) - . 

Sure we can simplify the signature of `RouteHandlerChain` if you think that are two many parameters with something like:
```java
public void handleRoute(RouteContext context) { ... }
```

but in this case in a regular handler I must write:
```java
public void handleRoute(RouteContext context) { 
    Request request = context.request();
    Response response = context.response();
    context.next();
}
```

but I don't see the value because we simplified the signature of method but we complicated the body. This `RouteContext` object maybe is valuable in the future if we want to put some business (methods) - of course is not ok to add more parameters in handleRoute and in this scenario we must create this context object (bucket).

gitblit:
If you don't move methods out of `Response` and `Request` and into `RouteContext` then I agree, there is no value and it makes `RouteHandler` implementations more complex.

[...]

PR #95, opened 2015-02-23
RFC: RouteContext

gitblit:
I've implemented a `RouteContext` as discussed in #91 and #92.

Basically, `RouteContext` is what used to be `RouteHandlerChain`.  I find this architecture to be cleaner and more flexible.

I eliminated the `Request` and `Response` factories and set those classes as final.  Here is why: I added a `RouteContextFactory` class to create the `RouteContext`.  Subclassing the `DefaultRouteContext` is much more useful than subclassing individual R&R classes as it allows you to make decisions on the complete request/response cycle, not just a piece of it.

I also eliminated the `Application` static singleton reference in `ControllerHandler`, which I find to be a big win.

You'll notice that I built-out `RouteContext` with some new methods which I have been using.  Obviously there is a bunch of overlap between this class and the R&R classes.  I think it would be ideal to make R&R simpler with more of the brains in `RouteContext`.

Lastly, `RouteContext` is an interface to allow mocking it in unit tests.

I'm sure it needs refinement, but I think this is the right track.
What do you think?


decebals:
I am OK with this PR. Thanks for your time and effort.

I am happy that you didn't remove `Request` and `Response` classes/concepts. These classes make `RouteContext` implementation more simple, short and clear. If you don't want to see these classes as abstraction of `HttpServletRequest/HttpServletResponse` you can see them eventually like utility classes and `RouteContext` implementation delegates to these classes.
Now to be honest I have a concern regarding the size of a RouteContext implementation (as number of lines and number of methods). I don't want to see a huge class with tones of methods.
Also I don't want to see ambiguous methods and to wonder if a method from `RequestContext` is related to the request or response. Fortunately the API for `RouteContext` proposed by you is very clear.

I see some minor problems but I think that we can merge this pull request if you have nothing to add to it.

Congratulations!

gitblit:
I'm done adding to this PR.

PR #99, opened 2015-03-02
Add support for naming routes to bring meaningful context to logs

gitblit:
Once your app gets large enough, logging by method and uri pattern is insufficient.

This PR adds the builder pattern to route registration with the first supported value as `Route` *name*.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/Application.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Application.java
@@ -188,62 +188,62 @@ public class Application {
         getRouter().setContextPath(contextPath);
     }
 
-    public void GET(StaticResourceHandler resourceHandler) {
+    public Route GET(StaticResourceHandler resourceHandler) {
         if (getRouter().uriPatternFor(resourceHandler.getClass()) != null) {
             throw new PippoRuntimeException("You may only register one route for {}",
                     resourceHandler.getClass().getSimpleName());
         }
         resourceHandler.setMimeTypes(mimeTypes);
         resourceHandler.setHttpCacheToolkit(httpCacheToolkit);
-        addRoute(resourceHandler.getUriPattern(), HttpConstants.Method.GET, resourceHandler);
+        return addRoute(resourceHandler.getUriPattern(), HttpConstants.Method.GET, resourceHandler);
     }
 
-    public void GET(String uriPattern, RouteHandler routeHandler) {
-        addRoute(uriPattern, HttpConstants.Method.GET, routeHandler);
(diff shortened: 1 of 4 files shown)


decebals:
Named routes make referring to routes so we can generate redirects or URLs more convenient.

gitblit:
That's a good idea and we should definitely do that. :+1:
e.g. `routeContext.redirect("login page")`.

decebals:
I don't know if we can use `routeContext.redirect("login page")` because already exists a method with this signature (`path` as parameter).
Right now we have (copied from readme):
```java
String url = getApplication().urlFor("/contacts/{id}", parameters);
```

Maybe it's simple something like `urlFor(route, parameters)` where `route` it's the route name or uriPattern (first I search route by name and if I didn't found a result I search route by uriPattern).
Also it's nice to have a `routeContext.redirectToRoute(route, paramaters)` (we can choose a short name for this method - `redirect`).

gitblit:
> already exists a method with this signature (path as parameter).

How about `void redirect(String pathOrName);`

And same thing for `urlFor(pathOrName, parameters)`.

decebals:
> How about void redirect(String pathOrName);
And same thing for urlFor(pathOrName, parameters).

Right now I see in `DefaultRouteContext`:
```java
public void redirect(String path) {
    response.redirectToContextPath(path);
}
```

and in `Response`:
```java
/**
 * Redirect the browser to a location which may be...
 * <ul>
 * <li>relative to the current request
 * <li>relative to the servlet container root (if location starts with '/')
 * <li>an absolute url
 * </ul>
 * If you want a context-relative redirect, use the {@link redirectToContextPath}
 * method.
 * <p>This method commits the response.</p>
 *
 * @param location
 *            Where to redirect
 */
public void redirect(String location) { ... }

/**
 * Redirects the browser to a path relative to the application context. For
 * example, redirectToContextPath("/contacts") might redirect the browser to
 * http://localhost/myContext/contacts
 * <p>This method commits the response.</p>
 *
 * @param path
 */
public void redirectToContextPath(String path) { ... }
```

We need something similar with:
```java
public void redirectToRoute(String nameOrUriPattern, Map<String, Object> parameters) { ... }
```

decebals:
What I want to say is that `redirect(String pathOrName)` is cryptic because we have mulptiple possibilities: `contextPath`, `location`, `routeName`, `routeUriPattern` (all multiplexed in `pathOrName` variable).

For `urlFor(pathOrName, parameters)` we have only `routeName` and `routeUriPattern` as significance for `pathOrName`. 

Maybe I'm wrong.

PR #100, opened 2015-03-04
Allow registration of custom ExceptionHandlers

gitblit:
This is a light-weight alternative to implementing a full `ErrorHandler` or subclassing `DefaultErrorHandler`.

The `ErrorHandler` is also an `ExceptionHandler` so if no registered `ExceptionHandler` is found for the exception, then the `ErrorHandler` implementation is used.  But the `ErrorHandler` is more than an `ExceptionHandler` because it is a registry of `ExceptionHandler` instances and it can generate status-code based results.

This is an important piece when integrating custom or library exceptions into a Pippo application.


decebals:
:+1:

PR #101, opened 2015-03-04
Move Initializers from PippoFilter to Application

gitblit:
Method names may need tweaking.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/Application.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Application.java
@@ -30,11 +30,21 @@ import ro.pippo.core.util.HttpCacheToolkit;
 import ro.pippo.core.util.MimeTypes;
 import ro.pippo.core.util.ServiceLocator;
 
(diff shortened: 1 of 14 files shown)


decebals:
It's ok. For method names I propose this schema:

```java
final void init() { ... };

final void destroy() { ... };

protected void onInit() { ... }

protected void onDestroy() { ... }
```

With this visibility schema we don't pollute the Application's API. My two cents.

gitblit:
`onInit()` and `onDestroy()` work for me.

The visibility decrease will hurt me for the short-run as I am not using `PippoFilter`.  I'm not sure I can re-align on `PippoFilter`.  Maybe.  I'll make the adjustments for the `on` methods but for now can we leave `init()` and `destroy()` public?

gitblit:
Refactor pushed.

decebals:
> The visibility decrease will hurt me for the short-run as I am not using PippoFilter.

You are not using PippoFilter? How?

Issue #96, opened 2015-02-24
Servlet container sticky points


gitblit:
I've been running Pippo on a couple containers and I have identified some under-the-hood issues.

1. `PippoFilter` is not necessarily loaded on startup of the container I think because it is a Filter instead of a Servlet.  This is an issue because I'd rather see Pippo complete it's initialization on startup rather than on the first request.  We might consider switching `PippoFilter` to be `PippoServlet`.  We don't lose anything AFAICS.
2. `PippoFilter` is called by the container for matching path requests (e.g. `/*`).  That means the response supplied to `PippoFilter` starts off at `200 OK`.  We then set `status(Integer.MAX_VALUE)` to indicate Pippo has not properly processed the response and we require that some handler in the chain to set the status to a valid value.<p />Unfortunately, some containers dislike `status(Integer.MAX_VALUE)` and other containers dislike `status(0)`. <p/>So one of the following needs to happen...
    - We identify an integer value that all containers are happy with: 1?  1000?
    - We set a flag in `Response` to indicate we have a set status (`boolean statusSet = false;`)<br/>**this is my vote**
    - We set a status int in `Response` that is used to set the status int in `HttpServletResponse` at the *end of the request*.

decebals:
> PippoFilter is not necessarily loaded on startup of the container I think because it is a Filter instead of a Servlet. This is an issue because I'd rather see Pippo complete it's initialization on startup rather than on the first request. We might consider switching PippoFilter to be PippoServlet. We don't lose anything AFAICS.

Personal I don't see a problem to initialize the application on the first request. The first request will take some more time. But if you think that a servlet it's teh best solution then I have nothing against in making this transition.
I found a link [1] about `Filter` vs `Servlet` for a web framework.

> - We identify an integer value that all containers are happy with: 1? 1000?
- We set a flag in Response to indicate we have a set status (boolean statusSet = false;)
this is my vote
- We set a status int in Response that is used to set the status int in HttpServletResponse at the end of the request.

I am ok with the solution 2 and 3. Maybe another solution is to set initial status to `NOT_FOUND` instead of `Integer.MAX_VALUE` or `0`.


[1] http://stackoverflow.com/questions/5110077/should-a-web-framework-be-a-filter-or-a-servlet

gitblit:
I pushed the change for the 404 a day or two ago.  Not sure if I am in love with that, but it is a low-cost choice.  Let's see how that goes.

[...]

PR #105, opened 2015-03-08
RFC: Add a Tjws WebServer implementation

gitblit:
Maybe worthwhile, maybe not.  I also checked out [Winstone](https://code.google.com/p/winstone) but it doesn't appear to have an API but it is an interesting micro-container.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/Response.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Response.java
@@ -46,8 +46,11 @@ public final class Response {
     private ContentTypeEngines contentTypeEngines;
     private TemplateEngine templateEngine;
     private Map<String, Object> locals;
+    private Map<String, String> headers;
     private Map<String, Cookie> cookies;
     private String contextPath;
+
+    private int status;
 
     public Response(HttpServletResponse httpServletResponse, Application application) {
         this.httpServletResponse = httpServletResponse;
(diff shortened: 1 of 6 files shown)


pippo-core/src/main/java/ro/pippo/core/Response.java
```diff
@@ -972,4 +950,21 @@ private void commit(CharSequence content) {
         }
     }
 
+    private void finalizeResponse() {
```
decebals: Maybe `finalize()` is better. I feel that `finalizeResponse` is redundant

decebals:
Seems ok to me. I have not tested to see that it works but I seem that changes are minor.

gitblit:
I've found a glitch that I didn't see before.  One fix is in the first commit of this PR (cache status).
This glitch was introduced in #96 because of the change from `status(Integer.MAX_VALUE)` to `response.notFound()`.  Not found is a valid status.  If we render a template without manually setting a status then we return a 404 with the properly rendered template.  `Response.send()` will set a `200 OK` if the current status is either 0 or MAX_VALUE.  I think we should merge this PR as it allows us to automatically handle status because we are delaying setting the status && we are avoiding container differences.

BTW, you mentioned an interest in async.  I think you should take a closer look at Undertow.  My understanding is that when you are running Pippo on Undertow you are automatically async.  I think the entire servlet API is run async.  Maybe not.  And my async, I mean Undertow has a small set of IO threads which delegate work to many worker threads.  I'm a little confused by the async buzzword so if you know more, please share.

Issue #115, opened 2015-04-21
RouteDispatcher ignores filterPath


decebals:
In `RouterDispatcher.onRouteDispatch()` the value for `requestPath` is not correct calculated. The correct calculation is in `PippoFilter.getRelativePath()` that by the way is not used anymore.

gitblit:
The issue was serving Pippo on a non-root filter path?

BTW, I would have named it `applicationPath` or `pippoPath` not `filterPath` - which implies use of `PippoFilter`.  Did `PippoServlet` also need to be updated?

decebals:
Yes. The issue occurs when you map `PippoFilter` (in `web.xml` in particular) on an non-root path (for example `/rest/*`). The `PippoServlet` is not affected by this issue.
Also I moved `setFilterPath()` from `PippoFilter` in `WebServer` because it's hard to transform a filterPath (for example /rest/*) in a canonical format (rest/).
In what file do you want to rename variable `filterPath`?

if you see a clearer resolution of the problem please let me know.

decebals:
You have right, my solution doesn't work for `PippoServlet` because `Response.setRelativePath()` is not called. I don't find a nice solution for both (Servlet and Filter). Before my commit we had duplication code (related to relative path in `PippoFilter` and `RouteDispatcher.onRouteDispatch()`). I will rethink my solution. Maybe a (Url)PathUtils class with static methods for different path types is a solution.

gitblit:
I think we rushed on implementation and release of 0.4.1 without thinking this completely through.
`PippoServlet` has this problem too - the analog would be *servletPath* instead of *filterPath*.

The issue gets larger when you consider templates....
```
${contextPath}/${pippoPath}/item/${id}
```

... and then, of course, reverse-routing.  At the moment, I think we are in poor shape for serving Pippo from a subpath of the context path.

decebals:
I released 0.4.1 for #113.

gitblit:
Sorry, I was mistaken.  When you get a chance, take a look at #116.  I think that covers all the cases.

decebals:
Resolved by #116.

decebals:
Are you OK with a new release (0.4.2) that should contain this fix?

gitblit:
Absolutely!

PR #116, opened 2015-04-24
Allow specifying a non-root Pippo filter or servlet path

gitblit:
How about this for #115?

Basic idea is to introduce an `applicationPath` which is composed of all path elements from server root to Pippo base path.  That requires adjusting Requests, Responses, TemplateEngines, etc.

The only thing I wasn't sure about was naming the template variable `${appPath}`.  If you want to change that we must be consistent across all engines.

Seems to work ok for me.  Tested with Pippo & Fathom.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/PippoFilter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/PippoFilter.java
@@ -233,30 +233,4 @@ public class PippoFilter implements Filter {
         }
     }
 
-    /**
-     * Returns a relative path to the filter path and context root.
-     */
-    private String getRelativePath(String contextPath, String path) {
-        String relativePath = path.substring(contextPath.length());
-
-        if (relativePath.length() > 0) {
-            relativePath = relativePath.substring(1);
-        }
-
-        if (!relativePath.startsWith(filterPath))  {
-            if (filterPath.equals(relativePath + "/"))  {
-                relativePath += "/";
-            }
-        }
-        if (relativePath.startsWith(filterPath)) {
-            relativePath = relativePath.substring(filterPath.length());
-        }
-
-        if (!relativePath.startsWith("/")) {
-            relativePath = "/" + relativePath;
-        }
-
-        return relativePath;
-    }
-
 }
--- a/pippo-core/src/main/java/ro/pippo/core/PippoFilter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/PippoFilter.java
@@ -78,13 +78,17 @@ public class PippoFilter implements Filter {
         try {
             String contextPath = StringUtils.addStart(filterConfig.getServletContext().getContextPath(), "/");
             application.getRouter().setContextPath(contextPath);
-            log.debug("Serving application on context path '{}'", contextPath);
 
             if (filterPath == null) {
                 initFilterPath(filterConfig);
             }
-            application.getRouter().setFilterPath(filterPath);
-            log.debug("The filter path is '{}'", filterPath);
+            String applicationPath = contextPath + filterPath;
+            application.getRouter().setApplicationPath(applicationPath);
+
+            if (!contextPath.equals(applicationPath)) {
+                log.debug("Context path is '{}'", contextPath);
+            }
+            log.debug("Serving application on path '{}'", applicationPath);
 
             log.debug("Initializing Route Dispatcher");
             routeDispatcher = new RouteDispatcher(application);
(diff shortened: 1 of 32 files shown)


decebals:
The PR is OK. I will made some inline comments just in case.

pippo-core/src/main/java/ro/pippo/core/PippoFilter.java
```diff
@@ -107,12 +111,16 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
         // TODO test for redirect
         // no redirect; process the request
 
-        String requestMethod = httpServletRequest.getMethod();
+        // create Request, Response objects
```
decebals: I see that `getRelativePath` method is not used anymore. Now we create the `Request`, `Response` instances in advance to use `Request.getApplicationRequestPath()` but are cases when the relativePath is ignored so we don't need the Request, Response instances.

pippo-core/src/main/java/ro/pippo/core/Request.java
```diff
@@ -277,6 +278,19 @@ public String getContextUri() {
     }
 
     /**
+     * Returns the uri relative to the application base uri.
+     *
+     * @return the uri relative to the application base url
+     */
+    public String getApplicationUri() {
```
decebals: I don't "feel" the method name. I must read the javadoc to understand what this method do. I don't know if `getRelativeUri` is a better name

pippo-core/src/main/java/ro/pippo/core/Request.java
```diff
@@ -436,21 +464,23 @@ public String getContextPath() {
...
+    public String getApplicationPath() {
+        return applicationPath;
     }
 
     /**
-     * Returns a relative path to the filter path and context root.
+     * Returns a request path relative to the Pippo application path.
      */
-    public String getRelativePath() {
+    public String getApplicationRequestPath() {
```
decebals: I have the same problem with the name. I don't know if it's not similar with `getApplicationUri()`.

decebals:
> The only thing I wasn't sure about was naming the template variable ${appPath}. If you want to change that we must be consistent across all engines.

For me it's OK. Maybe another option is `homePath` but it's not more clear than `appPath`.
I don't know if it's a good idea to add a template function similar with `${uriFor('/contact', <parameters_map>)}`

I see in `pippo-demo-crud` that in `CrudApplication.java` I use this snippet: 

```java
parameters.put("action", "save");
if (id > 0) {
    parameters.put("id", id);
}
routeContext.setLocal("editAction", getRouter().uriFor("/contact", parameters));
routeContext.setLocal("backAction", getRouter().uriFor("/contacts"));
routeContext.render("contact");
```

pippo-core/src/main/java/ro/pippo/core/PippoFilter.java
gitblit: I doubt there is a performance issue.  I'll remove `getRelativePath`.

gitblit:
I don't think we need a new template function for the `appPath`, a variable is enough.

Reverse-routing is already addressed by this PR so any `uriFor` calls would return the appropriate container-relative path.

On the get path method names.... I'm not in love with them either but now we have 3-tiers of paths...

1. relative to servlet container
2. relative to context path
3. relative to application path (filter or servlet path)

If you are serving a Pippo app on the embedded server on the root, then all these paths are the same.  If you are serving a Pippo app from a non-root WAR and are mounting Pippo on a non-root path within the WAR, then all three paths are different.  In this second scenario I can imagine I would need convenient methods to parse/generate urls for all three path reference points.

I'm open to suggestions on how to improve this.

decebals:
I pushed a little improvement. First, I fixed an issue in `PippoFilter` related to applicationPath computation. Also, I removed some unused methods form `Request` (for example getContext...()). I renamed `Request.applicationRequestPath` in simple `path`. I made some tests (with PippoFilter) with pippo-demo-basic running in an embedded  container (pippo-jetty) or in an external container (tomcat) and the results where good.

Is something that displeases you in my commit?

decebals:
Did you see other possible issues, something you do not like?

gitblit:
:+1: Looks good to merge.

I pushed a small bunch of fixes today which combined with this fix I think is grounds to push a maintenance release.

Maybe we should have a changelog?

And, BTW, have you seen [this](http://jitpack.io)? Pretty interesting service idea.  Especially when you can fork a repo, tag a release, and have them build & serve Maven artifacts for your own fork.  This has nothing to do with Pippo, just saw it and wanted to share.

decebals:
> Maybe we should have a changelog? 

Very good idea. I think that a changelog is a must have.
In the past I tried to use github [Releases](https://github.com/blog/1547-release-your-software) but without much success. Do you prefer the automatic approach via github Releases or the manual approach? If you prefer the manual approach who do you see the implementation (a changelog file, ...).

> And, BTW, have you seen this? Pretty interesting service idea. Especially when you can fork a repo, tag a release, and have them build & serve Maven artifacts for your own fork. This has nothing to do with Pippo, just saw it and wanted to share.

Thanks for sharing.

gitblit:
I generally don't use the Releases feature either.

For Gitblit I maintain a YAMLish changelog file.  I do that so I can build the site from the embedded documentation and generate an HTML changelog.  But that is a homegrown, complicated solution.

I recommend a simple file.  Probably Markdown.  It makes the most sense to have that file with the documentation sources.  I'm wondering if it's a good idea to keep pippo-site separate from pippo?

decebals:
Are you OK with a new release (0.4.2) that should contain this fix?

Issue #119, opened 2015-05-07
Possible error in handling query parameters which are arrays


balamaci:
Parameters as Arrays are written in the url by     
```
private String uriFor(PatternBinding binding, Map<String, Object> parameters) {
...
                query.append(parameterName).append("=").append(parameterValue.toString());
...
```

This generates some parameters like &tag=[phone,tablet]
which in turn gets deserialized by **ParameterValue** class:
```
    public List<String> toList(List<String> defaultValue) {
        if (isNull()) {
            return defaultValue;
        }

        if (values.length == 1) {
            return Arrays.asList(values[0].split(","));
        }
        return Arrays.asList(values);
    }
```
which results in a list of:
"[phone", "tablet]"

decebals:
I can replicate your problem with a tiny test.

```java
Map<String, Object> queryParameters = new HashMap<>();
queryParameters.put("tags", Arrays.asList("tablet", "monitor", "laptop"));
String result = router.uriFor("/test", queryParameters);
System.out.println("result = " + result);

List<String> list = new ParameterValue("[tablet, monitor, laptop]").toList();
System.out.println("list.size() = " + list.size());
for (int i = 0; i < list.size(); i++) {
    System.out.println("list.get(i) = " + list.get(i));
}
System.out.println("list = " + list);
```

The console looks like:
```
result = /test?tags=[tablet, monitor, laptop]
list.size() = 3
list.get(i) = [tablet
list.get(i) =  monitor
list.get(i) =  laptop]
list = [[tablet,  monitor,  laptop]]
```

I read that it's not a standard way of passing an array through a query string. I will investigate the problem.

gitblit:
I think the proper way to specify a multi-valued parameter is to specify multiple key-value pairs.
e.g. `/test?tags=tablet&tags=monitor&tags=laptop`

decebals:
Sure but I don't know how I can do this in a clean way.
I like:
```
Map<String, Object> parameters = new HashMap<>();
parameters.put("tags", Arrays.asList("tablet", "monitor", "laptop"));
String uri = router.uriFor("/test", parameters);
```
because is clean.

gitblit:
The issue is not the hashmap storage but the generated url.
This is non-functional code to give the general idea:

```java
public String uriFor(String path, Map parameters) {

    StringBuilder sb = new StringBuilder();
    sb.append(baseUrl + path);

    for (Entry entry : map.entrySet()) {
        Object value = entry.value();
[...]

Issue #35, opened 2014-12-15
Ajax


decebals:
Add support for Ajax requests

richdyang:
I don't understand what you mean. Ajax is more of front-end things. If you'd like the application to response as json, it's ok now. And returning javascript by template is also possible.
Ajax is really the difference in request manner and response message body. For server side, it's nothing to change.

decebals:
Your observation is correct. My idea is to supply a functional minimalist demo (based on jquery ajax). After this step I want to see if I can add something valuable in Pippo to simplify the development of Ajax applications (for example to create a pippo.js). One idea is to brings Pippo named routes to javascript.
It's only an idea with a low priority.

decebals:
I created a basic demo pippo-ajax. The ajax part is implemented using [intercooler.js](http://intercoolerjs.org/) - declarative ajax library.

gitblit:
That's a nifty little library!

gitblit:
I deployed a Bower Webjar for intercooler 0.4.10 so you shouldn't need to use the CDN.

<pre>
Deployed!
It will take a few hours for the Maven Central index to update but you should be
able to start using the Bower WebJar now.

GroupID = org.webjars.bower
ArtifactID = intercooler-js
Version = 0.4.10 
</pre>

decebals:
Nice. I see that it's very easy to write pippo-ajax application using intercoller (simple stuff). If you want, we can improve the pippo-demo-ajax module and add this demo on the master branch (and of course add a simple page `Ajax` on pippo-site).

gitblit:
> add this demo on the master branch

I think that would be a good idea.

decebals:
`<script src="${webjarsAt('intercooler-js/0.4.10/src/intercooler.js')}"></script>` is a little different than a standard `<script src="${webjarsAt('jquery/1.11.1/jquery.min.js')}"></script` (just an example) but it's a solution

gitblit:
I don't get to control the layout of the jar file.  That's part of the autoconversion from Bower to Webjar.

decebals:
Sure. I wrote above comment just to remember the correct syntax :smile:

decebals:
I will try to improve the demo with a CRUD

PR #128, opened 2015-05-29
Add support for 1-dimensional collections

gitblit:
This adds support for things like `List`, `Set`, and `TreeSet`.  The implementation covers Controllers and direct ParameterValue access from `RouteHandler`.  To compliment this feature, the `Request` will now recognize and assemble indexed-parameters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [#124]: Automatically encode the parameters values in DefaultRouter.uriFor()
 - [#126]: Maven quickstart archetype to build a small Pippo web application
 - [#128]: Added support for `Set`, `List`, and any other concrete Collection type
+- [#128]: Added support for array query/form parameters like `yada[0]`, `yada[1]`, & `yada[2]`
 
 #### Removed
 - Removed pippo-ioc module because it is no longer used anywhere
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [#121]: Added [pippo-jackson]
 - [#124]: Automatically encode the parameters values in DefaultRouter.uriFor()
 - [#126]: Maven quickstart archetype to build a small Pippo web application
+- [#128]: Added support for `Set`, `List`, and any other concrete Collection type
 
 #### Removed
 - Removed pippo-ioc module because it is no longer used anywhere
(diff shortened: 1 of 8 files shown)


decebals:
:+1: 
Maybe a little update of documentation (pippo-site) help more to understand this feature.

gitblit:
Definitely.  I will get that done - not today, though.

There are two halves to this feature

1. collections support
2. indexed parameter array support

Item 2 depends on 1 - but it is not the only way to use collections support.  If you supply multiple query parameters of the same name to a URL, that will result in a list of values which can take advantage of the collections too.  I'm not sure if you can simulate that with a form submit.  I think the last form input value of the same name will be used.  PHP and Django support item 2.  Maybe others too.

On a related front, #35, I've also been meaning to give [ajax-form](http://ajax-form.raynicholus.com/components/ajax-form/) a try.  Looks to be quite useful for declarative but modern forms.

gitblit:
This will need some additional work if we want to support Collections in Body types.  This PR is mainly for parameter types.

In #130 I demo'd POSTing a JSON array as the BODY.  That works because the list component type is a string.  Our body parsing API does not support arrays of complex type (i.e List<MyType>).  It could, though.

decebals:
Also I think that we have a bug.
For example #119 doesn't work.

```java
GET("/index", new RouteHandler() {

	@Override
		List<Integer> list = routeContext.getRequest().getParameter("list").toList(Integer.class);
		System.out.println("list = " + list);
	}

});

GET("/test", new RouteHandler() {

	@Override
		parameters.put("list", Arrays.asList(1, 2, 3));
		String uri = routeContext.uriFor("/index", parameters);
		System.out.println("uri = " + uri);
		routeContext.redirect(uri);
	}

});
```

Maybe I will create a quickstart for this.

gitblit:
That can probably be supported.  It is a convenient, albeit nonstandard? format.

decebals:
I don't care too much if the url query is standard or not for multi values parameter. The problem is that now the above snippet doesn't work.

> I think the proper way to specify a multi-valued parameter is to specify multiple key-value pairs.
e.g. /test?tags=tablet&tags=monitor&tags=laptop

I am ok with this approach, it is more standard.

Now, I see that I cannot upload the quickstart zip file so I will copy/paste the important part:

```java
GET("/index", new RouteHandler() {

	@Override
		List<Integer> list = routeContext.getRequest().getParameter("list").toList(Integer.class);
		System.out.println("list = " + list);
		routeContext.text().send(list);
	}

});

GET("/test", new RouteHandler() {

	@Override
		parameters.put("list", Arrays.asList(1, 2, 3));
		String uri = routeContext.uriFor("/index", parameters);
		System.out.println("uri = " + uri);
		routeContext.redirect(uri);
	}

});
``` 

In conclusion, for me the above issue is not important, we can leave it. It's a complex problem (for example I can put any object as parameter value) and I do not know if it's worth bother with it.

gitblit:
See #131.

PR #129, opened 2015-05-30
Allow html forms to specify _method which overrides the servlet method

gitblit:
This is the same useful behavior found in Rails and Django.  It allows for applying a more RESTful design to standard form processing.  To use it, specify a hidden input field named `_method` with a value of the preferred http method (e.g. `PATCH`, `PUT`, `DELETE`, etc).  Pippo will automatically route the POSTed request to the preferred handler.  The Collections controller example has been updated to use `PUT` as a demonstration.


gitblit:
Once we get this one merged, I'll have a similar follow-up for `_content_type` and `_content` which complete the ability of a form to exercise a RESTful API. See [this documentation](http://www.django-rest-framework.org/topics/browser-enhancements) for the motivation.

decebals:
thanks for the link with documentation

Issue #120, opened 2015-05-14
FileResourceHandler usage is not self evident


balamaci:
FileResourceHandler constructor is taking two parameters "url" and "directory"
https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/route/FileResourceHandler.java

making it confusing to use it like:
GET("/url", new FileResourceHandler("url", "directory")); 

it's intended to be used with:
GET(new FileResourceHandler("url", "directory"));

decebals:
I am ok with what you say. All static resource handlers are based on `StaticResourceHandler`. In the [constructor](https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/route/StaticResourceHandler.java#L47) of this class we create an uriPattern using the urlPath: `this.uriPattern = String.format("/%s/{%s: .*}", getNormalizedPath(urlPath), PATH_PARAMETER);` Method `getUriPattern` is used in [Application](https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/Application.java#L226). I don't know how we can eliminate this confusion.
All I see is a check in `GET("/url", RouteHandler)` in this form:
```
if (routeHandler instanceof StaticResourceHandler) {
    __error("Please use 'GET(StaticResourceHandler resourceHandler)'");
}
```

Maybe someone has other resolution.

gitblit:
I propose eliminating the duplicate handler check [from here](https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/Application.java#L221)

... and then something like this:
```java
public Route addWebjarResourceHandler() {
    return addWebjarResourceHandler("/webjars");
}

public Route addWebjarResourceHandler(String basePath) {
    return GET(new WebjarsResourceHandler(basePath));
}

public Route addPublicResourceHandler() {
    return addPublicResourceHandler("/public");
}

public Route addPublicResourceHandler(String basePath) {
    return GET(new PublicResourceHandler(basePath));
}

public Route addFileResourceHandler(String basePath, File directory) {
    return GET(new FileResourceHandler(basePath, directory));
}
```

gitblit:
Yeah, ok.  Maybe throwing `PippoRuntimeException` in `GET(String, RouteHandler)` is necessary to prevent it.

gitblit:
[...]

PR #136, opened 2015-06-08
#120

decebals:
What you say?

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/Application.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Application.java
@@ -258,53 +258,53 @@ public class Application {
     }
 
     /**
-     * It's a shortcut for {@link #addPublicRoute(String)} with parameter <code>"/public"</code>.
+     * It's a shortcut for {@link #addPublicResourceRoute(String)} with parameter <code>"/public"</code>.
      */
-    public Route addPublicRoute() {
-        return addPublicRoute("/public");
+    public Route addPublicResourceRoute() {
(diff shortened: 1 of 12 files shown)


pippo-core/src/main/java/ro/pippo/core/Application.java
```diff
@@ -376,6 +425,17 @@ public static Application get() {
         }
     }
 
+    private Route GET(StaticResourceHandler resourceHandler) {
+        if (getRouter().uriPatternFor(resourceHandler.getClass()) != null) {
+            throw new PippoRuntimeException("You may only register one route for {}",
```
gitblit: We should eliminate this check.  There are valid cases where you may want to register two different handlers of the same type.

pippo-core/src/main/java/ro/pippo/core/Application.java
```diff
@@ -258,6 +257,56 @@ public Route addRoute(String uriPattern, String requestMethod, RouteHandler rout
         return route;
     }
 
+    /**
+     * It's a shortcut for {@link #addPublicRoute(String)} with parameter <code>"/public"</code>.
+     */
+    public Route addPublicRoute() {
```
gitblit: How about `addPublicResourceRoute`  (and probably add *Resource* to the other methods)?  *addPublicRoute* sounds more like a security/authorization action.

gitblit: In Fathom I called this `addPublicResourceHandler`.  Maybe `Route` is better.

PR #140, opened 2015-06-11
Filter implementation

danjee:
Add a Tomcat runtime

### DIFF
--- a/pippo-tomcat/pom.xml
+++ b/pippo-tomcat/pom.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>pippo-parent</artifactId>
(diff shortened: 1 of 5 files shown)


gitblit:
Looks good.  Any particular reason for Tomcat 7 instead of Tomcat 8?

danjee:
I had a talk with Decebal and decided not to use the PippoFilter with EmptyServlet but instead to use the PippoServlet directly without PippoFilter. 
I will rework and make another pull request.
Also I will put tomcat 8 dependency. I haven't used Tomcat 8 so far, that was the reason.

decebals:
@gitblit What version you prefer between PippoFilter + EmptyServlet (I think that tomcat doesn't call the filter chain if you don't have a servlet on that context) vs PippoServlet?

gitblit:
I think the `PippoServlet` approach in #141 is cleaner - since this is an embedded `WebServer` it doesn't really need to use a `ServletFilter`.  Thinking about it some more, maybe all the `WebServer` implementations should take a `PippoServlet` instead of a `PippoFilter`.

danjee:
Maybe some rework is needed in `AbstractWebServer` by declaring some abstract methods so all implementations will follow a single pattern.
Also I think there is need for adding custom filters from client application for all implementations.

decebals:
> Thinking about it some more, maybe all the `WebServer` implementations should take a `PippoServlet` instead of a `PippoFilter`.

Make sense.

PR #148, opened 2015-06-14
Separate Query & Path parameters

gustavogalvan:


decebals:
It's ok. I will do some minor adjustments. Thanks.

pippo-core/src/main/java/ro/pippo/core/Request.java
```diff
@@ -128,6 +130,40 @@ public ParameterValue getParameter(String name) {
         return getAllParameters().get(name);
     }
 
+    /**
+     * Return all query parameters
+     */
+    public Map<String, ParameterValue> getQueryParameters() {
```
decebals: Here the name of method is not 100% true because you have Query Parameters + Post Parameters. I don't know if you can extract only the query parameters. Of course if method name is GET then the name is true. There are situations when you can have Query&Post parameters in the same request but this is another story and it's a particular case.

gustavogalvan:
beautiful

Issue #147, opened 2015-06-14
Web server tuning


gustavogalvan:
Make options (threads, buffer size, etc.) for each webserver accesible from settings files. This is specially usefull for Undertow, where speed varies with response size and buffer size needs to be configured.

decebals:
I pushed a possible implementation. Maybe it's not the best but it's work.

You must add this line in your ApplicationLauncher before `pippo.start();`
```java
((UndertowServer) pippo.getServer()).getSettings().setBufferSize(10 * 1024);
```

Any comment is welcome.

gustavogalvan:
Seems ok for me.
I downloaded the branch "server-settings" but doesn't compile jetty & undertow modules:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.5.1:compile (default-compile) on project pippo-jetty: Compilation failure: Compilation failure:
[ERROR] bootstrap class path not set in conjunction with -source 1.7
[ERROR] /home/gustavo/Proyectos/Pippo/pippo/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java:[44,50] error: type AbstractWebServer does not take parameters
[ERROR] /home/gustavo/Proyectos/Pippo/pippo/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java:[50,4] error: method does not override or implement a method from a supertype
[ERROR] /home/gustavo/Proyectos/Pippo/pippo/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java:[55,32] error: cannot find symbol

decebals:
I cannot reproduce your error. I compile the project from command line (maven) and also from IDE with both java 1.7 (17.0_25) and 1.8 (1.8.0_40) without errors. Maybe another user can confirm that it's an issue on compilation.

gustavogalvan:
Sorry ... my bad. Maven was using a pippo-core cached version. Compilation is OK, thanks !

decebals:
No problem. 
I see here some problems with the default settings.
For example at https://github.com/decebals/pippo/commit/4cf1353c7d168843b02e97a551ab4ce852b73716#diff-461ae40e92bd3b5964439e29c7c0fc47R109 I don't know if that check (bufferSize> 0) is the best solution. We don't know the default value so we can't know if the user made some custom settings or not.

gustavogalvan:
[...]

PR #149, opened 2015-06-15
a possible implementation for #147

decebals:


gitblit:
Looks good. Being able to tune from `PippoSettings` would be a plus: `undertow.bufferSize`.

PR #155, opened 2015-06-16
Add QueuedThreadPool to Jetty server

gustavogalvan:


decebals:
I prefer the approach from #149, #151. In few words:
- create a `JettySettings` (similar with [UndertowSettings](https://github.com/decebals/pippo/blob/master/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowSettings.java)) and put your settings in that file.
- change `JettyServer` to take into account the `JettySettings` class

I prefer each server with a custom settings `POJO` with some constants (properties names) and common settings in WebServerSettings.

What you say, make sense?

gustavogalvan:
definitely your approach is better.

decebals:
You come with another pull request or you want to create an issue?

gustavogalvan:
I'll try again with a new pull request, with your approach.

PR #156, opened 2015-06-17
Add more configurations to Jetty server

gustavogalvan:
Implements QueuedThreadPool.

In application.properties you can set:
jetty.maxThreads = 100
jetty.minThreads = 10
jetty.threadTimeout =30000

(threadTimeout in milliseconds)


decebals:
Excellent

decebals:
I have here an observation. In `JettySettings` we modified the default values of Jetty.
For example `minThreads` is 8, `maxThreads` is 200 and I don't see value for `threadTimeout` but in Jetty is named `idleTimeout`. Is it ok to override the default values for Jetty from Pippo, to put our defaults?

gustavogalvan:
You're right, defaults must be Jetty defaults.
Thanks

PR #160, opened 2015-06-19
Add Tomcat Settings

gustavogalvan:
Similar to Jetty and Undertow, but using HashMap because there are many options.
What do you think ?

This a first approach with some settings ... there are many more (see here: http://tomcat.apache.org/tomcat-8.0-doc/config/http.html). Thats why I used a HashMap.


gustavogalvan:
And there are an improvement: no need for default values ;)

gustavogalvan:
Closed because issue #166

PR #161, opened 2015-06-19
add route name as first criterion in urlFor

decebals:
I try to implements the idea from the comment https://github.com/decebals/pippo/pull/99#issuecomment-77233748

In few words I can add a route (in an hypothetical blog application) with:
```java
GET("/blogs/{year}/{month}/{day}/{title}", (routeContext) -> { routeContext.render("myTemplate)});
```

It's hard to create the reverse routing using the `uriPattern`:
```java
Map<String, Object> parameters = ...
routeContext.uriFor("/blogs/{year}/{month}/{day}/{title}", parameters);
```

Now you can add a `name` for a route and to use this name when you build the URL(reverse routing) to that route:
```java
GET("/blogs/{year}/{month}/{day}/{title}", (routeContext) -> { routeContext.render("myTemplate)}).named("blog");
```

The new cod becomes:
```java
Map<String, Object> parameters = ...
routeContext.uriFor("blog", parameters);
```

Advantages:
- it is often more descriptive than hard-coding the `uriPattern`
- it allows you to change `uriPattern` in one go, without having to remember to change URLs all over the place.

I will try to add the same approach for `redirect`.

What you say?

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouteContext.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouteContext.java
@@ -325,8 +325,8 @@ public class DefaultRouteContext implements RouteContext {
     }
 
     @Override
-    public String uriFor(String uriPattern, Map<String, Object> parameters) {
-        return application.getRouter().uriFor(uriPattern, parameters);
+    public String uriFor(String nameOrUriPattern, Map<String, Object> parameters) {
+        return application.getRouter().uriFor(nameOrUriPattern, parameters);
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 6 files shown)


gitblit:
I'm a big fan of the idea but It looks like I can only lookup by *name* once I have named a route.

decebals:
Yes, it's a bug. Thanks. I never thought that someone that someone wants to search by `uriPattern` when he has the shortcut with `name` :smile:

decebals:
I fixed it.

gitblit:
Well I doubt I would want to lookup by uri if I named it but if we advertise lookup by X or Y then it should work for both cases.

Looks good to me.  Merge it!

Issue #166, opened 2015-06-20
Web server settings improvements


gustavogalvan:
I'm suggesting a new approach for a more agile web server configuration, to avoid the need to recompile the server module every time a new option was "discovered". The truth is nobody in Pippo core has to be an expert for every web server, just only will know how configure it at Pippo integration.

From my point, all options could be dynamic at start time if it exists in application.properties, without need know all server's options.

This new approach must consider adding/changing options from outside classpath, for a "emergency cases", without need recompilation. For example, you use Undertow server and you see your system flooded of TIME_WAIT sockets (because higher concurrency), you want to set (Options.TCP_NODELAY, true) and (Options.REUSE_ADDRESSES, true), without loosing time.... 
1) adding to module (like I done):
        builder.setWorkerOption(Options.TCP_NODELAY, true);
        builder.setSocketOption(Options.TCP_NODELAY, true);
        builder.setSocketOption(Options.REUSE_ADDRESSES, true);
2) build the jar
3) modify your project
4) build your project
5) upload to server
6) restart the app

Adding or changing an option must be:
1) add / change the options
2) restart the app

What do you think ?

gustavogalvan:
Sorry for "must consider adding/changing options from outside classpath" ... this is already implemented by "-Dpippo.settings=/path/to/application.properties (flexible external location)". 
I must read the docs, I must read the docs, ...

gustavogalvan:
it's seems not easy, because each webserver has own strategy.

gustavogalvan:
How about this:
1) basic options (like now);
2) advanced (dynamic) options

In application.properties:
This advanced options will be unknown and we must know the variable type. The variable type could be inferred from text via regex, or declared explicity.

Then, in XxxxSettings:
1) get individual basic settings;
2) get all xxxx.advanced.* settings into HashMap

In XxxxServer:
1) apply basic settings (like now);
2) apply advanced settings (if not null)

Example:

application.properties:

# Tomcat
# Basic settings
tomcat.maxThreads = 100
tomcat.minSpareThreads = 10
# Adcanced settings
tomcat.server = "Pippo"
tomcat.socket.performanceConnectionTime = 1
tomcat.socket.performanceLatency = 2
tomcat.socket.performanceBandwidth = 0

Undertow is more complex because it provides three methods:
1) server options
2) worker options
3) socket options
But I don't believe that would change, so this is the particularity of this server.

PR #167, opened 2015-06-20
Added more settings to Undertow

gustavogalvan:


decebals:
For me it is too complicated with so many options. My idea is to supply only few options (I think the five existing options are sufficient) and if someone wants more she must override `Undetow.createServer(HttpHandler contextHandler)` and create a custom `Undertow` server.

gustavogalvan:
ok

decebals:
Thank you for your understanding.

gustavogalvan:
you're welcome.

I thought you was going to like it because with this scheme you do not have to worry about all options, nor need a method for each. Only declared options were applied, otherwise the server takes its default settings.

danjee:
I think it will be a pain keeping all this settings up to date with future Undertow releases. Also there are other servers that Pippo supports(Jetty, Tomcat, Tjws) and maybe other runtimes will be added in the future. If you have a complicated scheme for each you cannot keep the pace with the modifications.

gustavogalvan:
That's is a big true. If you like we can continue this debate in the issue #166 (where I'm posting a comment), or create a thread in google forum.

PR #169, opened 2015-06-21
Css preprocessor

danjee:
Add less4j and sass css compilers


gitblit:
Mixing two PRs together makes it hard to consider & review them independently.

Would you separate them?  It would also be nice to squash before opening the PR for less noise.

danjee:
OK, I will close it and open a new PR from a clean branch. I didn't know that I should use separate branches to split PR.

gitblit:
Thanks.  It makes it easier for us to review.  Conceptually, this would be a nice addition.

It would be worthwhile to consider how such integration should operate in PROD & TEST mode vs DEV mode.

PR #170, opened 2015-06-22
Added less and sass compilers

danjee:


pippo-sasscompiler/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>ro.pippo</groupId>
+            <artifactId>pippo-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-sass-compiler</artifactId>
```
decebals: Are many saas-compilers in java, [lesscss-java](https://github.com/marceloverdijk/lesscss-java) seems to be most popular. Why did you choose `vaadin-sass-compiler`?

decebals: Sorry. `lesscss-java` is less compiler and not a sass compiler.

danjee:
Vaadin seems a stable product that uses sass as css pre-processing. Gladly they pulled out the sass java compiler from vaadin sources and made it public. It's also an active project. 
The reason I choose less4j was that it's a pure java compiler (lesscss uses rhino for javascript)

decebals:
@danjee Thanks for your explanation.

decebals:
I have a dream to pack the `Less`, `Sass` functionalities in `StaticResourceHandler` via `LessResourceHandler` respective `SassResourceHandler` :smile: 
Also we can make `public` (now it's private) the method `Application.addStaticResourceRoute` and the code will become beautiful:
```java
addStaticResourceRoute(new LessResourceHandler(...));
addStaticResourceRoute(new SaasResourceHandler(...));
```

What you say? Is it possible?

gitblit:
That sounds like a good plan.

These handlers should also cache the generated CSS in PROD & TEST mode - no need to continuously recompile.

danjee:
@decebals I will make the modifications to use the handlers in that way. 
@gitblit I had an idea to keep a map with last modified time as key and the generated css as value but I think I will have a problem when there are multiple less/scss files and only a subset gets modified. I don't think I can check all of them for time modification. Do you have an idea on a different approach ?

gitblit:
These are supposed to be classpath resources, yes?  So in the production app the LESS/SASS is within the jar/war artifacts and they can not change so they only need to be compiled once.  In DEV mode we can recompile on every request.

This is similar to the approach we take with the `TemplateEngine` implementations.  In PROD & TEST mode the templates are compiled once and cached by the engines.  In DEV mode they are not cached at all and are therefore recompiled on every request.

I should think a `ConcurrentHashMap<String, String>` would be sufficient for caching, keyed by SASS/LESS resource name.  Am I overlooking something?

decebals:
Thanks!

pippo-less4j/src/main/java/ro/pippo/less4j/LessResourceHandler.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class LessResourceHandler extends StaticResourceHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LessResourceHandler.class);
+
+    private final String resourceBasePath;
+
+    private ConcurrentMap<String, String> sourceMap = new ConcurrentHashMap<>();
+
+    ThreadUnsafeLessCompiler compiler = new ThreadUnsafeLessCompiler();
```
gitblit: Hmm.  `ThreadUnsafeLessCompiler` and we are allowing concurrent access in DEV mode.  Doesn't that seem odd?

pippo-less4j/src/main/java/ro/pippo/less4j/LessResourceHandler.java
```diff
@@ -0,0 +1,85 @@
...
+            if (result == null) {
+                LessCompiler.CompilationResult compilationResult = compiler.compile(url);
+                for (LessCompiler.Problem warning : compilationResult.getWarnings()) {
+                    LOG.warn("Line: {}, Character: {}, Message: {} ", warning.getLine(), warning.getCharacter(), warning.getMessage());
+                }
+                result =  compilationResult.getCss();
+                sourceMap.put(content, result);
+            }
+            File file = File.createTempFile("tempfile", ".css");
+            FileWriter writer = new FileWriter(file);
```
gitblit: So for every request we are dumping the generated css from memory to a per-request temp file.
This should stream from memory - I don't see why we need to touch the filesystem at all.

pippo-sasscompiler/src/main/java/ro/pippo/sasscompiler/SassResourceHandler.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+            String content = scss.toString();
+            String result = sourceMap.get(content);
+            if (result == null) {
+                scss.compile(urlMode);
+                result = scss.printState();
+                sourceMap.put(content, result);
+            }
+            File file = File.createTempFile("tempfile", ".css");
+            FileWriter writer = new FileWriter(file);
```
gitblit: Dumping from memory to a per-request temp file and serving *that* is non-performant.  We need to redesign this to stream from memory.

gitblit:
@danjee Thanks for contributing this.  I plan to take advantage of it, looks to be quite useful!

pippo-sasscompiler/src/main/java/ro/pippo/sasscompiler/SassResourceHandler.java
decebals: @gitblit I agree with you. I wrote in a note:
> I think that we have some problems on multiple requests. Maybe I will introduce the Resource concept but for now it's safe to write in a file with a unique name (maybe via File.createTempFile).

The idea is that now we have `StaticResourceHandler.getResourceUrl(String resourcePath):URL` and the returned `URL` is used as parameter in `StaticResourceHandler.streamResource()`. In the Sass, Less handlers we have a String (css content) as result and not an URL. 
Sure the workaround with String->File->URL is a pain but it's a workable version that put some codes on table and we can/must improve this version.
Now, I think that we have two options:
- redesign `StaticResourceHandler`
- add `Resource` concept with some attributes(name, lastModified, inputStream, ...)

gitblit: Seems like the resource compilation handlers do not fit the design of `StaticResourceHandler` and we should not redesign the `StaticResourceHandler` hierarchy to accommodate them.  They just don't fit.

Either we need a new interface/type like `DynamicResourceHandler` or these should be normal handlers.

decebals: In this new `DynamicResourceHandler` we have the same code related to path parameter. Also we must to add a new method in Application `addDynamicResourceRoute` and I don't think that it's the best solution. Maybe an entire hierarchy with something like `ResourceHandler` as root is the best solution.

gitblit: Maybe so.  I'll sit back and wait for you and @danjee to propose a new scheme.  In the mean time I can sip my coffee and just be critical.  ;)

decebals: I don't have a new scheme. I try to evaluate (as concepts) some ideas. Nothing more. I don't want to be critical.

PR #172, opened 2015-06-23
Css preprocessor

danjee:


pippo-less4j/src/main/java/ro/pippo/less4j/LessResourceHandler.java
```diff
@@ -82,4 +85,10 @@ public String getResourceBasePath() {
         return resourceBasePath;
     }
 
+
+    public LessResourceHandler useCompressing(boolean compressing){
```
gitblit: That looks like a good idea.  I would automatically hook that to DEV mode.
```java
minify(!devMode);
```

danjee:
Unfortunately this feature is available only for less4j. sass-compiler will have this on the next release as I saw in github code. So you will need this only on PROD and TEST mode. You don't want the client application to have access to this feature ?

PR #173, opened 2015-06-23
Css preprocessor

danjee:
Rework by putting the less compiler as local variable for non thread safe access.

Add a fluent method for configuring minimized mode.

### DIFF
--- a/pippo-less4j/src/main/java/ro/pippo/less4j/LessResourceHandler.java
+++ b/pippo-less4j/src/main/java/ro/pippo/less4j/LessResourceHandler.java
@@ -45,6 +45,7 @@ public class LessResourceHandler extends ClasspathResourceHandler {
 
     public LessResourceHandler(String urlPath, String resourceBasePath) {
         super(urlPath, resourceBasePath);
+        this.compress = Application.get().getPippoSettings().isProd();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


danjee:
I'm not that pleased myself with the dumping of css result in a temporary file but I needed an URL for returning from getResourceUrl method. Is there a way to return a stream there where and put a content type on it ?

pippo-less4j/src/main/java/ro/pippo/less4j/LessResourceHandler.java
```diff
@@ -39,18 +39,12 @@
...
     private static final Logger log = LoggerFactory.getLogger(LessResourceHandler.class);
 
-    private final LessCompiler compiler;
-
     private Map<String, String> sourceMap = new ConcurrentHashMap<>(); // cache
 
-    public LessResourceHandler(String urlPath, String resourceBasePath) {
-        this(urlPath, resourceBasePath, new ThreadUnsafeLessCompiler());
-    }
+    private boolean compress;
```
decebals: An idea is to initialize this variable in constructor
```
this.compress = Application.get().getPippoSettings().isProd()
```

decebals:
The actual implementation of `StaticResourceHandler` works with URL. We try to find a solution, see https://github.com/decebals/pippo/pull/170#discussion_r33039265

decebals:
A comment like "initialize compress" for the last commit would have been more appropriate.

decebals:
This commit is a temporary solution. Thanks

PR #175, opened 2015-06-23
improvement resource handlers

decebals:
Attempt to solve the drawbacks found via #170. It's only a suggestion, any comments are welcome.


gitblit:
This looks good. I haven't run it yet. Instead of doing Application.get() in handle() you've got routeContext.getApplication ()

pippo-core/src/main/java/ro/pippo/core/route/UrlResourceHandler.java
```diff
@@ -0,0 +1,83 @@
...
+        if (url == null) {
+            routeContext.getResponse().notFound().commit();
+        } else {
+            streamResource(url, routeContext);
+        }
+    }
+
+    public abstract URL getResourceUrl(String resourcePath);
+
+    protected void streamResource(URL resourceUrl, RouteContext routeContext) {
```
decebals: I can remove this method and put the content inline directly in the method `handleResource`.  I don't know if someone wants to bypass this part with caching.

pippo-less4j/src/main/java/ro/pippo/less4j/LessResourceHandler.java
```diff
@@ -39,37 +36,30 @@
...
-        URL url = super.getResourceUrl(resourcePath);
-        if (url == null) {
-            return null;
-        }
-
+    protected void sendResource(URL resourceUrl, RouteContext routeContext) throws IOException {
         // clear cache for DEV mode
-        if (Application.get().getRuntimeMode().equals(RuntimeMode.DEV)) {
+        if (routeContext.getApplication().getPippoSettings().isDev()) {
             sourceMap.clear();
```
decebals: Is it better to touch the cache (`sourceMap`) only when I am  in `PROD` mode (no `sourceMap.get()`, `sourceMap.put()` in this mode)?

danjee: I agree with that. So to put records on that map only if in production mode. Do you want me to make the change?
Very nice work with the new RequestHandler :+1:

decebals: Sure, you can add a PR that fix this little issue, but without too much noise (with minimal impact) :smiley:

Issue #164, opened 2015-06-20
Webjars version


danjee:
A nice idea from wicket-bootstrap implementation of webjars was to use the `current` to get the latest version of the webjar, so if you modify the pom.xml to get other dependecy you shouldn't modify all the html or templates. Of course this should be optional and if the user wants to use the fixed version string the scripts should be loaded.

decebals:
I like the idea but I see some improvements (I don't know if it's technical possible or if add some value).
For example now we do:
```
<link href="${webjarsAt('bootstrap/3.3.1/css/bootstrap.min.css')}" rel="stylesheet">
<script src="${webjarsAt('jquery/1.11.1/jquery.min.js')}"></script>
```
You propose:
```
<link href="${webjarsAt('bootstrap/current/css/bootstrap.min.css')}" rel="stylesheet">
<script src="${webjarsAt('jquery/current/jquery.min.js')}"></script>
```
I prefer a very aggressive approach as:
```
<link href="${webjarsAt('bootstrap.min.css')}" rel="stylesheet">
<script src="${webjarsAt('jquery.min.js')}"></script>
```
or
```
<link href="${webjarsAt('css/bootstrap.min.css')}" rel="stylesheet">
<script src="${webjarsAt('js/jquery.min.js')}"></script>
```

And don't forget that we must keep `pippo-core` as slim as possible. If any implementation is a little big for core we must move the code in a new module (maybe it's not case of webjars).

gitblit:
Wow.  I'm not sure if we can safely get *that* aggressive with the WebJars handler.  The layout is consistent - but only up-to-a-point - especially when looking at the newer Bower WebJars.

Do you mean to search the classpath for a resource matching that name?  Could get expensive.

@danjee How does Wicket's *current* linking work?  Are they literally depending on `<version>current</version>` or are they interpreting *current* in their WebJar integration?

danjee:
They interpret the `current` word in java code.

decebals:
An idea is to use `WebJarAssetLocator` from `webjars-locator-core`
```
<dependency>
    <groupId>org.webjars</groupId>
    <artifactId>webjars-locator-core</artifactId>
    <version>0.25</version>
</dependency>
```
but again I don't know if it's ok (add some value) to complicate the current (simple) implementation of the  webjars in pippo.

gitblit:
Adding that jar would add quite a few [dependencies]  - Jackson & friends - to core and break one of our design requirements.

[...]

PR #180, opened 2015-06-27
Add support for Webjars `current` aliasing (#164)

gitblit:


gitblit:
We could also support an alias like: JQUERY/jquery.min.js to avoid a version spec altogether. That would be closer your aggressive proposal, @decebals.

decebals:
:+1: Very clean implementation.

decebals:
@gitblit If it's not tricky/complex from the implementation point of view, I think that will be great to have the the possibility to eliminate current word from declaration (`jquery/current/jquery.min.js` -> `jquery/jquery.min.js`). The decision is yours.

PR #183, opened 2015-06-27
Serialize and deserialize dates & times using ISO8601 in Jackson

gitblit:


decebals:
@gitblit It's a pleasure for me when see some unit tests. Shame on me that I don't add more such tests.

PR #185, opened 2015-06-27
Allow specification of Accept-Type by appending a registered suffix to the request path

gitblit:
This PR allows specification of the `Accept-Type` on the `Request` path.

    http://localhost:8338/contact
    http://localhost:8338/contact.json
    http://localhost:8338/contact.xml
    http://localhost:8338/contact.yaml

The path suffix is matched against the trailing pseudo-suffix of the registered `ContentTypeEngine`.

e.g. `application/json` = `json`
e.g. `application/x-yaml` = `yaml`

```java
GET("/contact(\\.(json|xml|yaml))?", new RouteHandler() {
    @Override
        Contact contact = new Contact()
            .setId(12345)
            .setName("John")
            .setPhone("0733434435")
            .setAddress("Sunflower Street, No. 6");
        routeContext.xml().negotiateContentType().send(contact);
    }
});
```

Aside from the verbose route path specification, the behavior is transparent to the end-user.  I have an idea on how to make this suffix support easier to use for route registration but it will require some rearchitecting to achieve.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/ContentTypeEngines.java
+++ b/pippo-core/src/main/java/ro/pippo/core/ContentTypeEngines.java
@@ -27,8 +27,8 @@ import java.util.TreeMap;
 
 /**
  * Container for registered content type engines. The main purpose of this
- * object is to gracefully handle mapping a complex content-type or accept
- * header to an engine.
+ * object is to gracefully handle mapping a suffix, complex content-type,
+ * or an accept header to an engine.
  *
  * @author James Moger
  */
@@ -38,8 +38,11 @@ public class ContentTypeEngines {
 
     private final Map<String, ContentTypeEngine> engines;
 
+    private final Map<String, ContentTypeEngine> suffixes;
+
     public ContentTypeEngines() {
         this.engines = new TreeMap<>();
+        this.suffixes = new TreeMap<>();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


gitblit:
@decebals I happen to think this change is particularly useful - although the route registration could stand improvement.  More on that if we can get this merged.

decebals:
I think that this feature can be useful for some people.

pippo-core/src/main/java/ro/pippo/core/ContentTypeEngines.java
```diff
@@ -93,25 +115,30 @@ public ContentTypeEngine registerContentTypeEngine(Class<? extends ContentTypeEn
...
      * text/html,application/xhtml+xml,application/xml;q=0.9,image/webp
      * </pre>
      *
-     * @param contentType
+     * @param contentTypeOrSuffix
      * @return null or the first matching content type engine
      */
-    public ContentTypeEngine getContentTypeEngine(String contentType) {
-        if (StringUtils.isNullOrEmpty(contentType)) {
+    public ContentTypeEngine getContentTypeEngine(String contentTypeOrSuffix) {
```
decebals: From implementation I see a permutation: `suffixOrContentType` instead of `contentTypeOrSuffix`. Am I wrong?

decebals:
Thanks

gitblit:
Cool.  Thanks for the merge.

decebals:
@gitblit Can you update [doc](http://www.pippo.ro/doc/routes.html) with this cool feature, please?

PR #186, opened 2015-06-27
Change Webjars path aliasing to only use the artifactId

gitblit:
This PR changes the Webjars handler version aliasing introduced in #180 to support implied request resource path version aliases like `jquery/` instead of requiring `jquery/current/` paths.

This means the first path segment (*artifactId*) will be substituted at runtime with *artifactId/version/* (e.g. `jquery/1.11.1/`) UNLESS the request resource path already starts with *artifactId/version/* (e.g. `jquery/1.11.1/`).


decebals:
:+1:

PR #188, opened 2015-07-01
Fix alias substitution in webjars paths

gitblit:
Found a nasty bug in the new webjars aliasing.

Example:
`${webjarsAt('foundation/js/foundation/foundation.alert.js')}`

will resolve to:
`/webjars/foundation/5.5.2/js/foundation/5.5.2/foundation.alert.js`

Note the double version substitution.  Iz bad.

This PR fixes that.  *I wouldn't be opposed to a quick 0.6.1.*

**Also.**
It should also be noted that our webjars aliasing may not work when the same version of a webjar is rebuilt with a `-n` suffix.

For example, font-awesome 4.3.0-1 is a rebuild of font-awesome 4.3.0 with an erroneous Maven dependency removed.  Inside the 4.3.0-1 jar the paths are still "4.3.0".  If we are brave we could strip the `-n` suffix but I honestly don't know if that will cause more harm than good.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/route/WebjarsResourceHandler.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/WebjarsResourceHandler.java
@@ -66,7 +66,7 @@ public class WebjarsResourceHandler extends ClasspathResourceHandler {
             // i.e. skip replacing first path segment of "/jquery/1.11.1/jquery.min.js"
             // BUT do replace first path segment of "jquery/jquery.min.js".
             if (!resourcePath.startsWith(artifactVersion)) {
-                String aliasedPath = resourcePath.replace(artifactPath, artifactVersion);
+                String aliasedPath = artifactVersion + resourcePath.substring(artifactPath.length());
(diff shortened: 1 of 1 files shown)


gitblit:
I have a small working patch for `-n` stripping.  Let me check with the Webjars artifacts to see if there are downsides to automatically stripping those.  Hold off on merge.

gitblit:
Ok, nevermind about `-n` stripping.  The several examples of rebuilds I looked at all matched correctly.

The font-awesome rebuilds seem to be the exception, not the rule and were perhaps they were hastily done.

decebals:
There's no rush. Where you are comfortable with the modifications I can merge this PR and I will create a new release (0.6.1).

gitblit:
Ok.  This looks good to merge.

gitblit:
For your reference, I filed webjars/font-awesome#14 and a rebuilt jar has been deployed which fixes the inconsistency.

decebals:
@gitblit Thanks!

PR #189, opened 2015-07-01
Register Json, Xml, and Yaml engines with pippo-jackson

gitblit:
Jackson supports reading/writing multiple formats using the same set of annotations.  In many cases you may not need any data annotations at all to read/write.

This PR refactors the Jackson integration to register Json, Xml, and Yaml engines.  It's a one-stop dependency for all three content types.

What do you think of this approach?

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/util/ClasspathUtils.java
+++ b/pippo-core/src/main/java/ro/pippo/core/util/ClasspathUtils.java
@@ -84,4 +84,19 @@ public class ClasspathUtils {
         return list;
     }
 
+    /**
+     * Returns true if the specified class can be found on the classpath.
+     *
+     * @param className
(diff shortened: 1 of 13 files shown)


decebals:
The idea is good but I see some problems when I want to use `pippo-jackson` for __ JSON__ and other pippo modules (`pippo-xstream`) for __XML__ (an example). In this case `JacksonInitializer` registers `JacksonXmlEngine` automatically and I cannot use another XmlEngine. 
Probably a solution is to declare `optional` the `jackson-dateformat-xml` and `jackson-dateformat-yaml` and to use some reflections to register or not `JacksonXmlEngine` and `JacksonYamlEngine`.

gitblit:
`optional` is a good compromise.  I'll make it so.

decebals:
:+1:

PR #192, opened 2015-07-02
Improve XStream usage

gitblit:


decebals:
In my opinion this PR resolves the problem for the moment. Will be nice if we have a possibility to set/create the xstream object from "outside".
I made a fast review related to `ContentTypeEngine` and I don't know if it's not better to:
- remove `ContentTypeEngine.init(Application)`
- rename `Application.registerConentTypeEngine(Class)` in `Application.addContentTypeEngine(ContentTypeEngine)`

in this situation, in `XstreamInitializer.init(Application)` I must create an instance of `XstreamEngine` using a constructor with parameter `Application` (if I need Application in my XstreamEngine).
Also it's necessary to add a property `xstream` in `XstreamEngine` with getter and setter to allow the creation of xstream object outside of XstreamEngine (maybe in Application.init with something like `(XstreamEngine) getContentTypeEngine(xml)).setXstream()`) ; I see that now the `xstream` object is recreated for each `toString`, `fromString`.
Also, we must check if we have the same problem for template engine. In some scenarios I want to retrieve the configuration/settings of that template engine (probably in Application.init) and to do some modifications (load templates from disk, enable cache, ...)

gitblit:
It's unlikely that the underlying library wrapped by a `ContentTypeEngine` is thread-safe so I don't think a `ContentTypeEngine` mutator makes much sense.  I agree that manipulating a `ContentTypeEngine` would be useful, but at that point it may be just as easy to create your own `ContentTypeEngine`.  I suppose for really common config options we could add some methods and fields.

To my mind, the value of `ContentTypeEngine.init(Application)` is we could setup the engine from `PippoSettings` (without having to add methods to the engine) or from some hook in our application subclass.

`TemplateEngines` are different than `ContentTypeEngines`.  `TemplateEngines` are thread-safe and we make a single instance of the engine so there it makes sense to expose accessors for the underlying library.

PR #194, opened 2015-07-10
a very basic (without validation) pippo-knockoutjs demo

decebals:
[KnockoutJS](http://knockoutjs.com) is one of my favorite javascript library. 

> Knockout is a JavaScript library that helps you to create rich, responsive display and editor user interfaces with a clean underlying data model. Any time you have sections of UI that update dynamically (e.g., changing depending on the user’s actions or when an external data source changes), KO can help you implement it more simply and maintainably.

I think that I can do more things related to `pippo-knockout `integration (for example I can generate from java the javascript for an entity - CRUD operations).

My only dilemma is if for these kind of demos we must create a separate repository (for example on github a new project `pippo-knockout`) or not. Maybe now we have too much modules that must be maintained.
What do you think?


gitblit:
Sounds awesome. I'm in the woods for a week. No laptop. A demo repo sounds like a good idea. And maybe pruning some of our basic demos.

decebals:
I am come back from vacation and I am ready for action :smile: 
If we move the demos in a new github repository (`pippo-demos` for example) I don't know how we can deal with the pippo dependencies (`pippo-core`, `pippo-jetty`, ...); we must add references to `RELEASE` or `SNAPSHOT`?
Also, is it worthwhile to make this change?

balamaci:
+1 demo should be moved. The nature of pippo with so many modules(that might even grow more) already can benefit for some separation. 
I think that using -SNAPSHOT as reference makes sense.

gitblit:
FYI I've found another little gem library that is pretty handy: http://www.listjs.com

decebals:
@gitblit Thanks for info.
I will try to create a new repository `pippo-demos` (are you ok with this name?) and I will move all demos in this repository. 
Maybe the next step is to create a new repository `pippo-contribs` (or with other name) and to let contributors to add pippo modules. I am not yet so sure that it's a good idea :smile:

decebals:
Moved to decebals/pippo-demo#1.

PR #199, opened 2015-07-23
Upgrade less4j version

danjee:


decebals:
Thanks! I am glad to see that you maintain the modules added by you.

PR #205, opened 2015-08-04
Allow multipart upload

### LINKED ISSUE #203: Support for multipart in Tomcat
In pippo-jetty we have support for multipart.

danjee:
Should fix #203


decebals:
Is it tested? If the response is false please test with https://github.com/decebals/pippo-demo/tree/master/pippo-demo-upload

danjee:
Yes, is tested. I will make a new test again with that demo

PR #210, opened 2015-08-24
remove "extends Assert" from each existent Test class

decebals:
I don't like the idea of static import in general but I think that we must do a concession for the test classes.
In pippo test classes we extended Assert class only to force the import but I think that it's bad for readability.

I used InteliiJ Idea so I used [this](http://stackoverflow.com/questions/5107023/add-favorite-methods-to-static-import-in-intellij) recommendation in my projects.


gitblit:
Wildcard static import vs subclassing... I don't have a strong opinion on the matter - at least in this context.

decebals:
@gitblit Thanks for the fast response, The main point is to remove Assert from the class declaration (`MyTest extends Assert`).

Issue #211, opened 2015-08-28
Add custom Filters, Extensions in PebbleEngine from Application


decebals:
The idea is that I want to extend Pebble template engine with custom Filters, Extensions.
For example I wish to use an extension built by me for my application. From Pebble {documentation ](http://www.mitchellbosecke.com/pebble/documentation/guide/extending-pebble) I see that it's very simple to do this:
```java
PebbleEngine engine = new PebbleEngine();
engine.addExtension(new MyCustomExtension());
```

but I don't have access to PebbleEngine from Application.

A possible use case could be:
```java
public class MyApplication extends Application {

   @Override
   protected void onInit() {
      PebbleEngine pebbleEngine = ((PebbleTemplateEngine) templateEngine).getPebbleEngine();
      engine.addExtension(new MyCustomExtension());

      // ... add routes
   }

}
```

I have the same problem with all template engines supported by Pippo (Freemarker was tested by me).
I don't know if it's not a good idea to make `TemplateEngine` generic like this:
```java 
public interface TemplateEngine<T> {

    // actual methods
    ...

    // new method
   public <T> get<a_good_name>();
    
}
```

Also I think that we can extend the concept to Servers (Jetty, Undertow) to "catch" the raw server and make some custom configuration (add a filter, or adjust a parameter). Maybe it's nice to have an interface 
```java
public interface X<T> {

    public <T> get<a_good_name>();

}
```

where X must be a nice and clear name.
With this interface I can modify TemplateEngine, WebServer to extends this interface.

What is your opinion?

PS. I have impression that the problem was discussed in other post but I didn't find the post :smile:

gitblit:
Yes. We should do this.  I have an immediate use-case for it - same as you.  I want to add filters or extensions to the Pebble instance.

decebals:
What variant? :smile: 
1. Add a simple `getPebbleEngine()` in PebbleTemplateEngine
2. Add generic to `TemplateEngine` + new method
3. Add interface X and modify `TemplateEngine` and `WebServer` signature

If it's 2. or 3. can you propose a nice name for the interface name and the method name? :smile:

gitblit:
It's gets a little tricky in the rest of the architecture when you start making everything generic.  Remember my proposal for subclassing router and routes and the immense amount of generic sigs in that PR?

I guess I would target 2 for now and see how that goes.

```java
interface TemplateEngine<X> {
   X getEngine();
}

[...]

PR #212, opened 2015-08-31
make TemplateEngine configurable (see #211)

decebals:
A proof of concept. See #211 for more information.

Conclusion:
- __Groovy__, we export `MarkupTemplateEngine` but my feeling is that we cannot modify some configuration parameters using this object
- __Jade__, we export `JadeConfiguration`, not the engine
- __Pebble__, we export `PebbleEngine`, perfect
- __Trimou__, we export `MustacheEngine` but all settings are done in a previous [step](https://github.com/decebals/pippo/blob/master/pippo-trimou/src/main/java/ro/pippo/trimou/TrimouTemplateEngine.java#L65) in `MustacheEngineBuilder`, my feeling is that we cannot modify some configuration parameters using this object (MustacheEngine)
- __Freemarker__, we export `Configuration`, not the engine
- __Velocity__, we export `Velocity` engine but  all settings are done in a previous [step](https://github.com/decebals/pippo/blob/master/pippo-velocity/src/main/java/ro/pippo/velocity/VelocityTemplateEngine.java#L68)

So, I don't have a perfect fit. I will try to dig more. Any idea is welcome.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/TemplateEngine.java
+++ b/pippo-core/src/main/java/ro/pippo/core/TemplateEngine.java
@@ -21,7 +21,7 @@ import java.util.Map;
 /**
  * @author Decebal Suiu
  */
-public interface TemplateEngine {
+public interface TemplateEngine<T> {
 
     public final static String DEFAULT_PATH_PREFIX = "/templates";
 
@@ -43,5 +43,7 @@ public interface TemplateEngine {
     public void renderString(String templateContent, Map<String, Object> model, Writer writer);
 
     public void renderResource(String templateName, Map<String, Object> model, Writer writer);
+
+    public T getEngine();
 
 }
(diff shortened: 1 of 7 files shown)


gitblit:
A perfect fit looks to be difficult.

Maybe a better alternative would be to make it easier to subclass a TemplateEngine and then manually assign it rather than trying to tweak whatever is discovered?

```java
public class TrimouTemplateEngine {
    @Override
    public void init(Applicaiton application) {
        // normal init stuff...

        // allow custom initialization
        init(builder);

        // create final engine
        engine = builder.build();
    }

    protected void init(MustacheEngineBuilder builder) {
    }
}

public class MyMoustache extends TrimouTemplateEngine {
    @Override
    protected void init(MustacheEngineBuilder builder) {
        builder.registerHelper("trimmer", new MyMoustacheTrimmer());
    }
}
```

With this we would do:

```java
Application.get().setTemplateEngine(new MyMoustacheEngine());
```

decebals:
Good idea. I will try your approach.

PR #213, opened 2015-09-01
make TemplateEngine configurable (see #211), reloaded

decebals:
Implementation based on [this](https://github.com/decebals/pippo/pull/212#issuecomment-136359578) comment.

I have a updated pippo-demo-template on pipe:
```java
public class PebbleDemo {

        // .peb is the default file extension
//        Pippo pippo = new Pippo(new TemplateApplication(new PebbleTemplateEngine(), "pebble/hello"));
        Pippo pippo = new Pippo(new TemplateApplication(new MyPebbleTemplateEngine(), "pebble/hello"));
        pippo.start();
    }

    public static class MyPebbleTemplateEngine extends PebbleTemplateEngine {

        @Override
        protected void init(PebbleEngine engine) {
            engine.addExtension(new AbstractExtension() {

                @Override
                public Map<String, Filter> getFilters() {
                    Map<String, Filter> filters = new HashMap<>();
                    filters.put("myupper", new MyUpperFilter());

                    return filters;
                }

            });
        }

    }

    public static class MyUpperFilter implements Filter {

        @Override
        public List<String> getArgumentNames() {
            return null;
        }

        @Override
        public Object apply(Object input, Map<String, Object> args){
            return (input != null) ? ((String) input).toUpperCase() : null;
        }

    }

}
```

### DIFF
--- a/pippo-freemarker/src/main/java/ro/pippo/freemarker/FreemarkerTemplateEngine.java
+++ b/pippo-freemarker/src/main/java/ro/pippo/freemarker/FreemarkerTemplateEngine.java
@@ -98,10 +98,9 @@ public class FreemarkerTemplateEngine implements TemplateEngine {
 
         webjarResourcesMethod = new WebjarsAtMethod(router);
         publicResourcesMethod = new PublicAtMethod(router);
-    }
 
-    public Configuration getConfiguration() {
-        return configuration;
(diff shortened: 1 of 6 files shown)


gitblit:
:+1: Looks good.

PR #215, opened 2015-09-09
Fix CSRF guard failure on content-types with specified charsets

gitblit:
The CSRF guard skips token validation for POST requests that are sent with charsets.
e.g. `Content-Type: application/x-www-form-urlencoded; charset=utf8`.

This is no bueno.  :-1:


decebals:
Thanks!

PR #218, opened 2015-09-29
Add method to finalize the Response and return the raw output stream

gitblit:
Pippo is an abstraction and convenience layer for the servlet API.  If you
need to implement custom streaming you can not use any of the Response
API when setting up headers, etc.  This change allows you to configure the
Response, as you normally would, and then obtain the raw output stream for
custom response streaming.


decebals:
Good. Thanks! I cannot merge this PR from github. A rebase is a must after previous PR. Please commit yourself.

PR #220, opened 2015-10-21
Allow use of UTF-8 in path parameter expressions

gitblit:
Add support for the `:alnum:`, `:alpha:`, `:ascii:`, `:digit:`, and `:xdigit:` POSIX character classes in URL path parameters.  These values are substituted for their Java equivalent when the Route PatternBinding is created by DefaultRouter.

Using the Java equivalent syntax *directly* creates complications with the existing route parameter matching expression due to nesting of curly braces.  This alternative is less invasive.


decebals:
:+1: Nice. It's for the first time when I read about  `POSIX character classes`.

PR #221, opened 2015-10-26
use standard java service loader mechanism via ServiceLocator and remove pippo.properties files

decebals:
The idea is to simplify the `Initializer` implementation and to use an unitary services/extensions mechanism in Pippo (we already use java `ServiceLoader` mechanism via `ServiceLocator` for `ContentTypeEngine`, `TemplateEngine`).


gitblit:
Looks good.

Issue #209, opened 2015-08-18
Exception handling for requests with arrays and very large numbers


lmicra:
Hello,

I'm getting the following exception with requests like:

//pippo-server/?t[14234234722234234234234123]

```
ERROR: UT005023: Exception handling request to /
java.lang.NumberFormatException: For input string: "14234234722234234234234123"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Integer.parseInt(Integer.java:583)
        at java.lang.Integer.parseInt(Integer.java:615)
        at ro.pippo.core.Request.initParameters(Request.java:148)
        at ro.pippo.core.Request.<init>(Request.java:74)
        at ro.pippo.core.RequestResponseFactory.createRequest(RequestResponseFactory.java:36)
        at ro.pippo.core.PippoFilter.doFilter(PippoFilter.java:117)
        at io.undertow.servlet.core.ManagedFilter.doFilter(ManagedFilter.java:60)
        at io.undertow.servlet.handlers.FilterHandler$FilterChainImpl.doFilter(FilterHandler.java:132)
        at io.undertow.servlet.handlers.FilterHandler.handleRequest(FilterHandler.java:85)
        at io.undertow.servlet.handlers.security.ServletSecurityRoleHandler.handleRequest(ServletSecurityRoleHandler.java:62)
        at io.undertow.servlet.handlers.ServletDispatchingHandler.handleRequest(ServletDispatchingHandler.java:36)
        at io.undertow.servlet.handlers.security.SSLInformationAssociationHandler.handleRequest(SSLInformationAssociationHandler.java:131)
        at io.undertow.servlet.handlers.security.ServletAuthenticationCallHandler.handleRequest(ServletAuthenticationCallHandler.java:57)
        at io.undertow.server.handlers.PredicateHandler.handleRequest(PredicateHandler.java:43)
        at io.undertow.security.handlers.AbstractConfidentialityHandler.handleRequest(AbstractConfidentialityHandler.java:46)
        at io.undertow.servlet.handlers.security.ServletConfidentialityConstraintHandler.handleRequest(ServletConfidentialityConstraintHandler.java:64)
        at io.undertow.security.handlers.AuthenticationMechanismsHandler.handleRequest(AuthenticationMechanismsHandler.java:58)
        at io.undertow.servlet.handlers.security.CachedAuthenticatedSessionHandler.handleRequest(CachedAuthenticatedSessionHandler.java:72)
[...]

PR #222, opened 2015-11-15
try-catch, defaults to string param

lmicra:
A pull-request to implement a try-catch for the issue reported in #209


decebals:
Thanks!

Issue #138, opened 2015-06-10
Move all handlers from core in a new package handler?


decebals:
A new package `core.handler` and move all handlers (*Handler.java, route.*Handler.java) in this package.

gitblit:
I agree, the *core* package is a little fat.  Another alternative would be moving `*Handler.java` into *core.route*?  We already have that and they are directly related - a *Route* has a url, method, and a Handler.  That would add 5-6 classes to *core.route* perhaps making it too fat.  If we a create a new package that moves 5-6 classes out of *core* and 5-6 classes out of *core.route*.  I'm fine with whatever you think is best.

Unrelated, but looking at the *Route* constructor I'm struck by feeling the signature is backwards:

```java
public Route(String uriPattern, String requestMethod, RouteHandler routeHandler)
```

I think everywhere else we specify the order as METHOD URL (.e.g. GET /webjars/jquery...)

PR #227, opened 2015-11-29
Cosmetics for Route class

decebals:
This pull request comes with three changes:
- Change order of requestMethod, uriPattern (Route constructor) according to #138
- Change order of requestMethod, uriPattern (Route.findRoutes)
- Add static factory methods for GET, POST, ... in Route class

I think that code is more clear. The signature for the Route constructor was changed but I think that it's not a big problem because nobody uses the Route constructor directly.

I will add some javadoc if you are OK with these modifications.


gitblit:
| nobody uses the Route constructor directly

... except me... but your changes look good.

:+1:

Issue #228, opened 2015-11-29
Versioning public resources


gitblit:
It would be nice to somehow inject a version code into the public resource urls.  This would force browsers to always fetch updated CSS or JS files, for example.

decebals:
Done

PR #229, opened 2015-11-30
Versioning public resources (#228)

decebals:
I propose you an implementation for #228. My intention was to find a simple solution with minimal modifications of the existing code.
Also I wished a solution that is independent by chosen template engine.
The current implementation uses lastModified property. For the moment all static resources are versionable.

Show me the code :)

Template: `hello.peb`
```html
<head>
    <link href="{{ publicAt('css/style.css') }}" rel="stylesheet">
</head>
```

Rendered html:
```html
<head>
    <link href="/public/css/style-1448911263314.css" rel="stylesheet">
</head>
```

Pippo debug console:
```
21:20:36 [qtp1963387170-13] DEBUG ro.pippo.core.util.ClasspathUtils - Located 'templates/hello.peb' in the context classpath
21:20:36 [qtp1963387170-13] DEBUG ro.pippo.core.route.DefaultRouter - Inject version in resource path: 'css/style.css' => 'css/style-1448911218869.css'
21:20:36 [qtp1963387170-13] DEBUG ro.pippo.core.route.RouteDispatcher - Returned status code 200 for GET '/template'
21:20:36 [qtp1963387170-15] DEBUG ro.pippo.core.PippoFilter - Request GET '/public/css/style-1448911218869.css'
21:20:36 [qtp1963387170-15] DEBUG ro.pippo.core.route.DefaultRouter - Found 1 route matches for GET '/public/css/style-1448911218869.css'
21:20:36 [qtp1963387170-15] DEBUG ro.pippo.core.route.DefaultRouteContext - Executing handler for GET '/public/{path: .*}'
21:20:36 [qtp1963387170-15] DEBUG ro.pippo.core.route.UrlResourceHandler - Remove version from resource path: 'css/style-1448911218869.css' => 'css/style.css'
21:20:36 [qtp1963387170-15] DEBUG ro.pippo.core.route.UrlResourceHandler - Streaming as resource 'file:/C:/Users/decebal/work/myproject/target/classes/public/css/style.css'
21:20:36 [qtp1963387170-15] DEBUG ro.pippo.core.route.RouteDispatcher - Returned status code 200 for GET '/public/css/style-1448911218869.css'
21:21:10 [qtp1963387170-16] DEBUG ro.pippo.core.PippoFilter - Request GET '/template'
21:21:10 [qtp1963387170-16] DEBUG ro.pippo.core.route.DefaultRouter - Found 1 route matches for GET '/template'
21:21:10 [qtp1963387170-16] DEBUG ro.pippo.core.route.DefaultRouteContext - Executing handler for GET '/template'
21:21:10 [qtp1963387170-16] DEBUG ro.pippo.core.util.ClasspathUtils - Located 'templates/hello.peb' in the context classpath
21:21:10 [qtp1963387170-16] DEBUG ro.pippo.core.route.DefaultRouter - Inject version in resource path: 'css/style.css' => 'css/style-1448911263314.css'
21:21:10 [qtp1963387170-16] DEBUG ro.pippo.core.route.RouteDispatcher - Returned status code 200 for GET '/template'
21:21:10 [qtp1963387170-19] DEBUG ro.pippo.core.PippoFilter - Request GET '/public/css/style-1448911263314.css'
21:21:10 [qtp1963387170-19] DEBUG ro.pippo.core.route.DefaultRouter - Found 1 route matches for GET '/public/css/style-1448911263314.css'
21:21:10 [qtp1963387170-19] DEBUG ro.pippo.core.route.DefaultRouteContext - Executing handler for GET '/public/{path: .*}'
21:21:10 [qtp1963387170-19] DEBUG ro.pippo.core.route.UrlResourceHandler - Remove version from resource path: 'css/style-1448911263314.css' => 'css/style.css'
21:21:10 [qtp1963387170-19] DEBUG ro.pippo.core.route.UrlResourceHandler - Streaming as resource 'file:/C:/Users/decebal/work/myproject/target/classes/public/css/style.css'
21:21:10 [qtp1963387170-19] DEBUG ro.pippo.core.route.RouteDispatcher - Returned status code 200 for GET '/public/css/style-1448911263314.css'
```

So, the implementation works but probably we should make some small adjustments (better performance maybe ?!).

Ideas for this feature:
- disable versioning
- select a different strategy (the current strategy is based on lastModified property)

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
@@ -407,7 +407,8 @@ public class DefaultRouter implements Router {
     }
 
     private String uriFor(PatternBinding binding, Map<String, Object> parameters) {
-        String uri = binding.getRoute().getUriPattern();
+        Route route = binding.getRoute();
+        boolean isResourceRoute = UrlResourceHandler.class.isAssignableFrom(route.getRouteHandler().getClass());
 
         List<String> parameterNames = binding.getParameterNames();
         if (!parameters.keySet().containsAll(parameterNames)) {
@@ -416,16 +417,22 @@ public class DefaultRouter implements Router {
 
         Map<String, Object> queryParameters = new HashMap<>(parameters.size());
 
+        String uri = route.getUriPattern();
+
         for (Entry<String, Object> parameterPair : parameters.entrySet()) {
             boolean foundAsPathParameter = false;
 
             StringBuffer sb = new StringBuffer();
-            String buffer = String.format(VARIABLE_PART_PATTERN_WITH_PLACEHOLDER, parameterPair.getKey());
-
-            Pattern pattern = Pattern.compile(buffer);
+            String regex = String.format(VARIABLE_PART_PATTERN_WITH_PLACEHOLDER, parameterPair.getKey());
+            Pattern pattern = Pattern.compile(regex);
             Matcher matcher = pattern.matcher(uri);
             while (matcher.find()) {
                 String pathValue = parameterPair.getValue().toString();
+                if (isResourceRoute && ResourceHandler.PATH_PARAMETER.equals(parameterPair.getKey())) {
+                    String versionedResourcePath = ((UrlResourceHandler) route.getRouteHandler()).injectVersion(pathValue);
+                    log.debug("Inject version in resource path: '{}' => '{}'", pathValue, versionedResourcePath);
+                    pathValue = versionedResourcePath;
+                }
                 matcher.appendReplacement(sb, pathValue);
                 foundAsPathParameter = true;
             }
@@ -447,11 +454,11 @@ public class DefaultRouter implements Router {
                 Entry<String, Object> parameterEntry = iterator.next();
                 String parameterName = parameterEntry.getKey();
                 Object parameterValue = parameterEntry.getValue();
-                String encodedParameterValue = null;
+                String encodedParameterValue;
                 try {
                     encodedParameterValue = URLEncoder.encode(parameterValue.toString(), PippoConstants.UTF8);
                 } catch (UnsupportedEncodingException e) {
-                    throw new PippoRuntimeException("Cannot encode the parameter value '" + parameterValue.toString() + "'", e);
+                    throw new PippoRuntimeException("Cannot encode the parameter value '{}'", e, parameterValue.toString());
                 }
                 query.append(parameterName).append("=").append(encodedParameterValue);
 
--- a/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
@@ -408,7 +408,7 @@ public class DefaultRouter implements Router {
 
     private String uriFor(PatternBinding binding, Map<String, Object> parameters) {
         Route route = binding.getRoute();
-        boolean isResourceRoute = UrlResourceHandler.class.isAssignableFrom(route.getRouteHandler().getClass());
+        boolean isResourceRoute = ResourceHandler.class.isAssignableFrom(route.getRouteHandler().getClass());
 
         List<String> parameterNames = binding.getParameterNames();
         if (!parameters.keySet().containsAll(parameterNames)) {
@@ -429,8 +429,7 @@ public class DefaultRouter implements Router {
             while (matcher.find()) {
                 String pathValue = parameterPair.getValue().toString();
                 if (isResourceRoute && ResourceHandler.PATH_PARAMETER.equals(parameterPair.getKey())) {
-                    String versionedResourcePath = ((UrlResourceHandler) route.getRouteHandler()).injectVersion(pathValue);
-                    pathValue = versionedResourcePath;
+                    pathValue = ((ResourceHandler) route.getRouteHandler()).injectVersion(pathValue);
                 }
                 matcher.appendReplacement(sb, pathValue);
                 foundAsPathParameter = true;
--- a/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
@@ -429,7 +429,10 @@ public class DefaultRouter implements Router {
             while (matcher.find()) {
                 String pathValue = parameterPair.getValue().toString();
                 if (isResourceRoute && ResourceHandler.PATH_PARAMETER.equals(parameterPair.getKey())) {
-                    pathValue = ((ResourceHandler) route.getRouteHandler()).injectVersion(pathValue);
+                    ResourceHandler resourceHandler = (ResourceHandler) route.getRouteHandler();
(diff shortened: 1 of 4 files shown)


pippo-core/src/main/java/ro/pippo/core/route/UrlResourceHandler.java
```diff
@@ -49,6 +59,50 @@ public final void handleResource(String resourcePath, RouteContext routeContext)
 
     public abstract URL getResourceUrl(String resourcePath);
 
+    /**
+     * Inject version fragment.
+     */
+    public String injectVersion(String resourcePath) {
+        URL resourceUrl = getResourceUrl(resourcePath);
+        try {
+            long lastModified = resourceUrl.openConnection().getLastModified();
```
gitblit: If we are in `PROD` mode this should cache.

pippo-core/src/main/java/ro/pippo/core/route/UrlResourceHandler.java
```diff
@@ -33,12 +35,20 @@
 
     private static final Logger log = LoggerFactory.getLogger(UrlResourceHandler.class);
 
+    private static final Pattern VERSION_PATTERN = Pattern.compile("-[0-9]+\\.");
```
gitblit: Maybe instead of using `-` which is a fairly common character, perhaps we should use a different delimiter like `~` or another alternative.

pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
```diff
@@ -416,16 +417,22 @@ private String uriFor(PatternBinding binding, Map<String, Object> parameters) {
...
-
-            Pattern pattern = Pattern.compile(buffer);
+            String regex = String.format(VARIABLE_PART_PATTERN_WITH_PLACEHOLDER, parameterPair.getKey());
+            Pattern pattern = Pattern.compile(regex);
             Matcher matcher = pattern.matcher(uri);
             while (matcher.find()) {
                 String pathValue = parameterPair.getValue().toString();
+                if (isResourceRoute && ResourceHandler.PATH_PARAMETER.equals(parameterPair.getKey())) {
+                    String versionedResourcePath = ((UrlResourceHandler) route.getRouteHandler()).injectVersion(pathValue);
+                    log.debug("Inject version in resource path: '{}' => '{}'", pathValue, versionedResourcePath);
```
gitblit: After we finalize a design perhaps this should be `trace`.  Pippo is a little verbose when logging request processing.

pippo-core/src/main/java/ro/pippo/core/route/UrlResourceHandler.java
```diff
@@ -33,12 +35,20 @@
...
+
     public UrlResourceHandler(String urlPath) {
         super(urlPath);
     }
 
     @Override
     public final void handleResource(String resourcePath, RouteContext routeContext) {
+        String unversionedResourcePath = removeVersion(resourcePath);
+        if (!unversionedResourcePath.equals(resourcePath)) {
+            log.debug("Remove version from resource path: '{}' => '{}'", resourcePath, unversionedResourcePath);
```
gitblit: Same note here on log level.

gitblit:
Nice job!  I'll give it a go later today.

pippo-core/src/main/java/ro/pippo/core/route/UrlResourceHandler.java
decebals: Another idea is `-ver-`, for example `style-ver-165476554.css`

pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
decebals: Sure, or we can remove this log line if you considerate it unimportant

gitblit:
I moved the logging and switched to your proposed `-ver-` syntax.  I'm a little wary of adding the overhead of versioning to everything (webjars, sass, less, etc) but I can see the value.  To cutdown on the file IO of looking up mod times I adjusted the version scheme to be hexadecimal and the webjars handler is now hashing the cached aliases as the resource version (the alias includes the maven version).

decebals:
I added a `versioned` in `ResourceHandler` so we can disable cache on some specific resources types (for example webjars).
The default value for `versioned` flag is true. I think that we don't want versioning for webjars but I may be wrong. 
What you say?

gitblit:
I think it makes more sense to log injection and removal in the same class rather than split across classes.

 I think the webjar version hascode will only be injected if you are using the version-less webjars syntax. In this case I think it's OK to inject the hashcode.

decebals:
> I think it makes more sense to log injection and removal in the same class rather than split across classes.

I made the modification.

Now, I don't understand why we need hexadecimal in `UrlResourceHandler.getResourceVersion()`. It's a time consuming with `Long.toHexString` and we can return very easy a string from `lastModified` long.

gitblit:
It doesn't need to be; I'm just biased toward symmetry.

In looking at `Long.toString()` and `Long.toHexString()`, I think it's a wash on time consumption but if you prefer `toString()` that's fine with me.

decebals:
I intend to merge this pull request if you don't have another observation.

pippo-core/src/main/java/ro/pippo/core/route/UrlResourceHandler.java
```diff
@@ -49,6 +53,59 @@ public final void handleResource(String resourcePath, RouteContext routeContext)
 
     public abstract URL getResourceUrl(String resourcePath);
 
+    protected String getResourceVersion(String resourcePath) {
+        URL resourceUrl = getResourceUrl(resourcePath);
+        try {
+            return Long.toString(resourceUrl.openConnection().getLastModified());
+        } catch (IOException e) {
+            throw new PippoRuntimeException("Failed to read lastModified property for {}", e, resourceUrl);
```
gitblit: If we are trying to access a resource that does not exist I think we're going to throw a `500` for the request instead of a `404` because we are throwing this exception.

We might consider swallowing the `IOException` and returning `null` if we can't `getLastModified`.  It won't hurt anything, I don't think, and it will allow more understandable results (i.e. 404).

pippo-core/src/main/java/ro/pippo/core/route/UrlResourceHandler.java
```diff
@@ -49,6 +53,63 @@ public final void handleResource(String resourcePath, RouteContext routeContext)
...
     public abstract URL getResourceUrl(String resourcePath);
 
+    protected String getResourceVersion(String resourcePath) {
+        String version = null;
+
+        URL resourceUrl = getResourceUrl(resourcePath);
+        try {
+            version =  Long.toString(resourceUrl.openConnection().getLastModified());
+        } catch (IOException e) {
+            log.error("Failed to read lastModified property for {}", resourceUrl, e);
```
gitblit: Now that I think about this some more this will probably throw an NPE when it can't locate the resource which will still trigger a `500`.  We may want to trap that.

decebals:
@gitblit You have the freedom to change the code.

PR #231, opened 2015-12-14
Add PippoTest

decebals:
Add the possibility for an easy testing. 
Add `PippoTest` class in a new module `pippo-test`, with a simple API, based on rest-assured [1].
I took a look on fathom-test-tools.

How to use:

1. Add the dependency to `pippo-test` in your `pom.xml`
```
<dependency>
    <groupId>ro.pippo</groupId>
    <artifactId>pippo-test</artifactId>
    <version>${pippo.version}</version>
    <scope>test</scope>
</dependency>
```

2. Add a test in your application
```java
public class RoutesTest extends PippoTest {

    @Override
    public Application createApplication() {
        return new MyApplication();
    }

    @Test
    public void testIndex() {
        when().
            get("/").
        then().
            statusCode(200);
    }

}
```

In the current implementation, `PippoTest` creates a new `Pippo` instance for each Test. Sure, another option is to create a `Pippo` instance for all tests but I prefer to "push" this functionality in the actual class (maybe with a flag in `setUp` method that tell me if the instances was already created - I don't know if it's possible or if it's safe) and not to enter a new class with `@BeforeClass` and `@AfterClass`.

This is only a draft and I invite you at discussion.

[1] https://github.com/jayway/rest-assured

### DIFF
--- a/pippo-test/pom.xml
+++ b/pippo-test/pom.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <parent>
+        <groupId>ro.pippo</groupId>
+        <artifactId>pippo-parent</artifactId>
+        <version>0.8.0-SNAPSHOT</version>
+    </parent>
+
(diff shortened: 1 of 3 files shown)


gitblit:
Hi DS.  It's going to take me a little while to be able to assist on this branch.  I have a looming project deadline - a project based on Pippo+Fathom, actually.  I use the Fathom testing code everyday and it's nice.  So this is definitely good stuff to upstream.

decebals:
A little documentation [here](https://groups.google.com/forum/#!topic/pippo-java/N-3GMVsXZW8).

PR #238, opened 2016-01-25
Gracefully handle parameter values composed of whitespace

gitblit:
In real-world usage of Pippo I found users would "reset" a field by
overwriting a value (such as a date) with whitespace.  This would
cause exceptions when interpreting the empty values as dates.

Most ParameterValue.toSomething methods should be able to cope with
empty whitespace strings as inputs.


decebals:
Thanks!

PR #239, opened 2016-01-25
Implement DirectoryResourceHandler for serving external directories

gitblit:


decebals:
First time when I saw this PR, I must recognize that I was a little confused :smile: A combination between Welcome file and Directory files listing. Maybe it's easy to resume in an word (Directory) what does this resource handler. But now I am comfortable and I see the utility of this class in some cases.
I see some improvements (chain configuration - return `this` at the end of each `set` methods, maybe add the possibility to add a `FileFilter`) in the future.

Thanks!

gitblit:
It's a starting point and definitely could use some additional features.  I needed this capability for one of my projects.  This is similar in concept to Apache serving a WWW-ROOT but with optional template generation.

`DirectoryResourceHandler` supports `HEAD` but I didn't automatically register `HEAD` routes in `Application`, so that might be a future improvement.

decebals:
After a short review, I have some doubt.
First, `DirectoryResourceHandler` doesn't extend `ResourceHandler` so it's a little unclear for me. In my acceptation, a resource handler is a route handler that deals/responds with files (virtual or concrete files, from classpath or from file system). I don't know if this handler is a resource handler in above context.
If it's a resource handler then `DirectoryResourceHandler` must extend `ResourceHandler` or probably `FileResourceHandler`. Otherwise we must rename this class to something like `DirectoryHandler`.
All I want to do is to keep things a little bit organized in my mind.

gitblit:
`DirectoryHandler` is fine.  Subclassing `ResourceHandler` is fine too if we push the definition of the regexes down from `ResourceHandler`.  The regex needs for a directory url is different from that of a file url.  The existing `ResourceHandler` regex requires we match a file.

decebals:
I made the modifications. Also I removed `addDirectory...` methods from Application because I think that these methods are rare used and I wish to keep the Application class as simple/clean as possible (from API point of view).

gitblit:
Great. :+1:

PR #241, opened 2016-01-28
Transparently support Java 8 -parameters names for controllers

gitblit:
We've discussed this a couple times.  Now that we are on Java 8 this is a no-brainer.

This change allows you to define a controller method like this:

```java
public class ContactsController extends Controller {
    public void uriFor(int id) {
    }
}
```

instead of like this:

```java
public class ContactsController extends Controller {
    public void uriFor(@Param("id") int id) {
    }
}
```

Of course, to have the name automatically embedded in your code **requires** modifying your project compiler flags to specify `-parameters` to javac.  Your IDE needs this specified also, separately AFAICT.

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <version>3.1</version>
    <configuration>
        <source>1.8</source>
        <target>1.8</target>
        <compilerArguments>
            <parameters/>
        </compilerArguments>
    </configuration>
</plugin>
```

If you do not compile with this flag then `isNamePresent` will return false and the method will return `null`, just like before.  Which means this change works transparently.

### DIFF
--- a/pippo-controller/src/main/java/ro/pippo/controller/DefaultControllerHandler.java
+++ b/pippo-controller/src/main/java/ro/pippo/controller/DefaultControllerHandler.java
@@ -27,10 +27,12 @@ import ro.pippo.core.util.StringUtils;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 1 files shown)


decebals:
For me it's an improvement to eliminate the value field from `@Param`.

For example this code:
```java
public class ContactsController extends Controller {
    public void uriFor(@Param("id") int id) {
    }
}
```

is a little simple/clean in this form:
```java
public class ContactsController extends Controller {
    public void uriFor(@Param int id) {
    }
}
```

Do you say that I can eliminate the `@Param` annotation entirely? 
So
```java
public class ContactsController extends Controller {
    public void uriFor(int id) {
    }
}
```
is enough?

gitblit:
> Do you say that I can eliminate the @Param annotation entirely?

TL;DR: No, we can not `git rm Param.java` - if that is what you mean.
FULL:
1. If you do NOT specify the `-parameters` flag then you must use `@Param`.  If you do specify `-parameters` then you no longer need to use `@Param`.
2. There may be cases where you want the path or query parameter name to be different than the method argument name.  If we remove the `@Param` annotation then we can not handle that case.

decebals:
I know that we cannot remove the Param.java file. My question was if I can remove `@Param` from the controller method definition and your response is clear.

decebals:
Thanks!

PR #242, opened 2016-02-03
Stop manually maintaining pippo-pebble META-INF/services files

gitblit:
I discovered this [slick little APT compiler](http://metainf-services.kohsuke.org) which would get us out of the business of manually maintaining `META-INF/services/` files.

Interested?

### DIFF
--- a/pippo-pebble/pom.xml
+++ b/pippo-pebble/pom.xml
@@ -40,5 +40,11 @@
 			<artifactId>prettytime</artifactId>
 		</dependency>
 
+        <dependency>
+            <groupId>org.kohsuke.metainf-services</groupId>
+            <artifactId>metainf-services</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
 	</dependencies>
 </project>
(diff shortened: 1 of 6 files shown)


decebals:
It's a solution. I implemented somethig similar in the last PF4J version, [ServiceProviderExtensionStorage](https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/ro/fortsoft/pf4j/processor/ServiceProviderExtensionStorage.java)

decebals:
> metainf-services.jar is only necessary during the compilation, so specify it as an optional dependency to avoid having the jar in the runtime. 

This is good.

decebals:
Can you supply a most generic PR (to use `@MetaInfServices` in all Pippo modules)? 
If you are busy I can try to resolve this task.

gitblit:
> Can you supply a most generic PR (to use @MetaInfServices in all Pippo modules)?

It seems you are willing to merge, so, yes, I can do that.  This was a *test-the-waters* PR without going through all that effort.

Did you want to split out your implementation from PF4J or use the one in this PR?

decebals:
I am OK with your propose. We can use the implementation from this PR (external dependency because we have this dependency only to runtime).

PR #243, opened 2016-02-04
Stop manually maintaining all META-INF/services files

gitblit:
There are two options for how to use this tool:

```java
@MetaInfServices
public class MyEngine implements ContentTypeEngine {
```

and

```java
@MetaInfServices(ContentTypeEngine.class)
public class MyEngine implements ContentTypeEngine {
```

The less verbose style does not work for classes with multiple interfaces or hierarchies.  So this PR currently uses the second more verbose style for everything.

If you wish to use the less verbose style, where possible, I can refine.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,11 @@
                 <artifactId>prettytime</artifactId>
                 <version>3.2.5.Final</version>
             </dependency>
+            <dependency>
+                <groupId>org.kohsuke.metainf-services</groupId>
+                <artifactId>metainf-services</artifactId>
+                <version>1.1</version>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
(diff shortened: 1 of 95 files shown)


pom.xml
```diff
@@ -151,6 +151,11 @@
                 <artifactId>prettytime</artifactId>
                 <version>3.2.5.Final</version>
             </dependency>
+            <dependency>
+                <groupId>org.kohsuke.metainf-services</groupId>
+                <artifactId>metainf-services</artifactId>
+                <version>1.1</version>
```
decebals: I'm OK with this version. I see that the last version is 1.7 (http://mvnrepository.com/artifact/org.kohsuke.metainf-services/metainf-services)

decebals:
Please add an entry in changelog. Thanks!

PR #246, opened 2016-02-11
Setup RestAssurred with ContentTypeEngines

gitblit:
This allows you to re-use the server's content type engines from your restful tests.

```java
Item item = given().accept(JSON).when().get("/api/v1/items/{id}", 1).as(Item.class);
assertEquals("Item id does not match", 1, item.getId());
assertEquals("Item name does not match", "Apples", item.getName());
```


decebals:
Very good. Thanks!

PR #249, opened 2016-02-12
use Java7 Regex Named Groups fixed PathRegexBuilder wrong pathParameter bug

ScienJus:
> I have actually gotten this to work properly now.

> I had already thought of adding a /user part to the URL but I really like the /{username} URL so I wanted to try to make that work first.

> I have also tried to exclude certain routes like /login using a PathRegexBuilder. This works but there is a bug in getting the path parameter.

> Let's say I go to the route "/Username". When I try to get the username path parameter using getPathParameter("username") it should give me the value "Username", right? Instead it gives me the value "/Username" (note that the forwardslash is included in the value). 

> I have fixed this by using the String.replace function, but this is essentially a bug I think. It only happens when I use the PathRegexBuilder.

from [google group][1]

i tried pippo `0.8.0` and `0.9.0-SNAPSHOT`, it really happened, because the regex from `PathRegexBuilder.build` had extra groups, then `matcher.group(index)` get wrong group.

like:

```
// new PathRegexBuilder().includes("/{username}").excludes("/login").build();
("/aaa" =~ /(?=^(\/([^\/]*)).*)^(?!\/login).+/).each { x, y, z ->
    println x   // group(1) -> /aaa wrong
    println y   // group(2) -> /aaa wrong
    println z   // group(3) -> aaa  right
}
```

a easy ways to fix it is use Regex Named Groups in Java 7 (we just supported java 8, not 6, right?), we can give every group a name, just change `(.*?)` to `(?<name>.*?)`, then we can use `matcher.group(name)` to get the group. i think it will be more useful.

[1]: https://groups.google.com/forum/#!topic/pippo-java/7GHaR7JTnuU

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
@@ -53,7 +53,7 @@ public class DefaultRouter implements Router {
     private static final Pattern PATTERN_FOR_VARIABLE_PARTS_OF_ROUTE = Pattern.compile("\\{(.*?)(:\\s(.*?))?\\}");
 
     // This regex matches everything in between path slashes.
-    private static final String VARIABLE_ROUTES_DEFAULT_REGEX = "([^/]*)";
+    private static final String VARIABLE_ROUTES_DEFAULT_REGEX = "(?<%s>[^/]*)";
 
     // This regex works for both {myParam} AND {myParam: .*}
     private static final String VARIABLE_PART_PATTERN_WITH_PLACEHOLDER = "\\{(%s)(:\\s([^}]*))?\\}";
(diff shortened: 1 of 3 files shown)


ScienJus:
> I have actually gotten this to work properly now.

> I had already thought of adding a /user part to the URL but I really like the /{username} URL so I wanted to try to make that work first.

> I have also tried to exclude certain routes like /login using a PathRegexBuilder. This works but there is a bug in getting the path parameter.

> Let's say I go to the route "/Username". When I try to get the username path parameter using getPathParameter("username") it should give me the value "Username", right? Instead it gives me the value "/Username" (note that the forwardslash is included in the value). 

> I have fixed this by using the String.replace function, but this is essentially a bug I think. It only happens when I use the PathRegexBuilder.

from [google group][1]

i tried pippo `0.8.0` and `0.9.0-SNAPSHOT`, it really happened, because the regex from `PathRegexBuilder.build` had extra groups, then `matcher.group(index)` get wrong group.

like:

```
// new PathRegexBuilder().includes("/{username}").excludes("/login").build();
("/aaa" =~ /(?=^(\/([^\/]*)).*)^(?!\/login).+/).each { x, y, z ->
    println x   // group(1) -> /aaa wrong
    println y   // group(2) -> /aaa wrong
    println z   // group(3) -> aaa  right
}
```

a easy ways to fix it is use Regex Named Groups in Java 7 (we just supported java 8, not 6, right?), we can give every group a name, just change `(.*?)` to `(?<name>.*?)`, then we can use `matcher.group(name)` to get the group. i think it will be more useful.

[1]: https://groups.google.com/forum/#!topic/pippo-java/7GHaR7JTnuU

ScienJus:
if you think it's ok, i will add more testing to make sure it is working properly

decebals:
I like your initiative :smile: Please add more tests and I will merge this PR.

ScienJus:
@decebals finished

decebals:
Great. Thanks!

Issue #237, opened 2016-01-24
Add declarative approach 


decebals:
See https://github.com/gitblit/fathom/issues/8

decebals:
I created a [branch](https://github.com/decebals/pippo/tree/improve_contoller). I have a little issue in DefaultControllerHandler (I don't find a good solution to remove the Guice Injector so the code doesn't pass the compile phase). 
For now I don't have enough time to finish the work so any help is welcome.

gitblit:
I'll try to take a look later this week.

decebals:
Thanks James! It's not an easy task but it's important.

danjee:
This will be interesting

decebals:
I will try to finish work to this feature in next days.

decebals:
I will try to [move forward](https://groups.google.com/forum/#!topic/pippo-java/rclBFE-bc-U) this feature. I hope that this time will be lucky :smile:

decebals:
Done by #341.

PR #250, opened 2016-02-14
route group in pippo-controller

ScienJus:
i think pippo controller do not need nested group, so it will be simpler like:

```
ControllerRouteGroup group = new ControllerRouteGroup("/users", UserController.class);
group.GET("{id}", 'get');
group.PUT("{id}", 'modify');

new ControllerApplication.addRouteGroup(group);
```

but there is one problem that we can't return a Route after`group.GET("{id}", 'get')`, because we need `controllerHandlerFactory` to create the `RouteHandler`.

so i created a new method named `newRouteGroup` in `CountrollerApllication` and give `controllerHandlerFactory` to `ControllerRouteGroup` in this method, then we can return the route we created. like:

```
ControllerApplication application = new ControllerApplication();

ControllerRouteGroup group = application.newRouteGroup("/users", UserController.class);
Route route = group.GET("{id}", 'get').named('getUser');
group.PUT("{id}", 'modify');

application.addRouteGroup(group);
```

absolutely, `inGroup` method can not be supported in this route.

then, i was thinking about why we need `Route.inGroup`, it can be instead of`RouteGroup.addRoute`, and if we removed this method, we can calculate absolute uri pattern before create a new Route, like:

```
//Application.newRouteGroup
public RouteGroup newRouteGroup(String uriPattern) {
    return new RouteGroup(uriPattern, this);    // <- give application to group
}

//RouteGroup.GET
public Route GET(String uriPattern, RouteHandler routeHandler) {
    return this.application.GET(getAbsoluteUriPattern(uriPattern), routeHandler);
}

//RouteGroup.getAbsoluteUriPattern
private String getAbsoluteUriPattern(String routeUriPattern) {
    return this.groupUriPattern + routeUriPattern
}
```

after that we also don't need `Route.group` and `RouteGroup.routes/parent/children`, all of them was exists just because `Route.inGroup` method. Do we really need them?

i will add tests in pippo-controller after we have decided how to make group api simpler and more useful.

### DIFF
--- a/pippo-controller/src/main/java/ro/pippo/controller/ControllerApplication.java
+++ b/pippo-controller/src/main/java/ro/pippo/controller/ControllerApplication.java
@@ -75,6 +75,10 @@ public class ControllerApplication extends Application {
         return ALL(uriPattern, createRouteHandler(controllerClass, methodName));
     }
 
+    public ControllerRouteGroup newRouteGroup(String uriPattern, Class<? extends Controller> controllerClass) {
+        return new ControllerRouteGroup(uriPattern, controllerClass, this);
+    }
+
     public ControllerHandlerFactory getControllerHandlerFactory() {
         if (controllerHandlerFactory == null) {
             ControllerHandlerFactory factory = ServiceLocator.locate(ControllerHandlerFactory.class);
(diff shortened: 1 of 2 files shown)


decebals:
I don't know what to say about the route groups in pippo-controller. I did something in [this](https://github.com/decebals/pippo/issues/237) direction (for the moment I cannot compile project yet to create a remote branch on github) and if we choose the declarative approach I don't know if we need the variant with route group.
I will try to take a look on this subject (route group in controller). I will come with news.

ScienJus:
declarative approach is great!

decebals:
This PR doesn't make sense in the context of new controller concept that is available from version 1.2.0.

PR #252, opened 2016-02-16
route group

ScienJus:
some description:

1, why use `setGroupUriPattern` and `absoluteUriPattern`? because i think maybe some people will repeated add route to group, like:

```
RouteGroup group1 = new RouteGroup("/users");
RouteGroup group2 = new RouteGroup("/posts");
Route route = Route.POST("like", emptyRouteHandler);
group1.addRoute(route);
group2.addRoute(route);  // it will raises an exception
```

2, why `parent-group` shared routes to `child-group`

```
public RouteGroup(RouteGroup parent, String uriPattern) {
    if (parent != null) {
        this.uriPattern = StringUtils.concatUriPattern(parent.getUriPattern(), uriPattern);
        this.routes = parent.getRoutes();   // parent-children groups shared routes
    } else {
        this.uriPattern = uriPattern;
        this.routes = new ArrayList<>();
    }
}
```

also because maybe some people forget add the top-group to application… like:

```
RouteGroup group = new RouteGroup("/users");
RouteGroup child = new RouteGroup(group, "{id}");
Route route = Route.POST("like", emptyRouteHandler);
child.addRoute(route);

// forget it
//router.addRouteGroup(group);

router.addRouteGroup(child);
```

what do you think?

there will be some problems if used improperly in group.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
@@ -195,11 +195,13 @@ public class DefaultRouter implements Router {
     @Override
     public void addRouteGroup(RouteGroup routeGroup) {
         routeGroup.getRoutes().forEach(this::addRoute);
+        routeGroup.getChildren().forEach(this::addRouteGroup);
     }
 
     @Override
     public void removeRouteGroup(RouteGroup routeGroup) {
         routeGroup.getRoutes().forEach(this::removeRoute);
+        routeGroup.getChildren().forEach(this::removeRouteGroup);
     }
 
     @Override
--- a/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/DefaultRouter.java
@@ -195,13 +195,11 @@ public class DefaultRouter implements Router {
     @Override
     public void addRouteGroup(RouteGroup routeGroup) {
         routeGroup.getRoutes().forEach(this::addRoute);
-        routeGroup.getChildren().forEach(this::addRouteGroup);
     }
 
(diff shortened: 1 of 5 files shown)


decebals:
> 1, why use setGroupUriPattern and absoluteUriPattern? 

It's OK for now. Other variant is to clone this route and add the clone but it's another story.

> 2, why parent-group shared routes to child-group

Here it's a little complicated and a source for trouble. I'm not happy with this. I would simplify this aspect, so if you forget to add the group you don't have routes from that group, it's fair.

ScienJus:
> so if you forget to add the group you don't have routes from that group, it's fair.

ok, fixed it

decebals:
Thanks! I will make some minor modifications.

decebals:
I feel that I found two issues in `DefaultRouter.java`

```java
    @Override
    public void addRouteGroup(RouteGroup routeGroup) {
        routeGroup.getRoutes().forEach(this::addRoute);
        routeGroup.getChildren().forEach(this::addRouteGroup); // <-- REMOVE line
    }

    @Override
    public void removeRouteGroup(RouteGroup routeGroup) {
        routeGroup.getRoutes().forEach(this::removeRoute);
        routeGroup.getChildren().forEach(this::removeRouteGroup); // <-- REMOVE line
    }
```

We must delete two lines. @ScienJus Is my observation correct?

ScienJus:
no, this is a recursive method to make sure when we add the group routes we will also add children group's routes in this method, like:

```
RouteGroup parent = new RouteGroup("/users");
RouteGroup child = new RouteGroup(parent, "/posts");
Route route = child.GET('{id}', emptyRouteHandler);

router.add(parent);  // router.routes.size -> 1, but if we remove this line, router.routes will be empty because route is in child not parent
```

PR #255, opened 2016-02-27
fixed the bug when underline in path parameter name

ScienJus:
in https://github.com/decebals/pippo/pull/249 i try to use regex named group to fix wrong path parameter value bug. but there will have a new problem that underline is not supported in group name. for example if the uri pattern like `/{user_id}`, there will be a `PatternSyntaxException`:

```
java.util.regex.PatternSyntaxException: named capturing group is missing trailing '>' near index 8
/(?<user_id>[^/]+)
```

so i have to change group name from parameter name to parameter index like `/(?<param1>[^/]+)`(`/(?<1>[^/]+)` is not supported too).


decebals:
Thanks!

decebals:
By the way, I prefer dash instead of underscore. Probable because some time ago I read something similar with [this](http://blog.woorank.com/2013/04/underscores-in-urls-why-are-they-not-recommended/) article.

Issue #256, opened 2016-03-01
Date Param Support


icirellik:
I'm trying to pass a Date in as a parameter to a controller and can't seem to find any examples on how this should work. After some digging through the code I noticed a second annotation `@ParamPattern`, do I need to set a pattern for to make this work?

``` java
public void getHistory(@Param("since") Date since) {
  ...
}
```

decebals:
For now, `@ParamPattern` is used only inside of the entity object (POJO).

For example (it's a pseudocode):
```java
class User {

    @Param
    String name;

    @Param
    @ParamPattern
    Date since;

    // getters and setters

}
```

and now you can use
```java
class MyController extends Controller {

    void createUser(@Form User user) {
        userService.createUser(user);
    }

}
```

We try to improve the controller module with #237 (probably in the next version you will have this improvement). For now I will review the solution proposed by @ScienJus in #257.

icirellik:
This is exactly what I needed, thanks. This works for now.

decebals:
Is it the problem solved? Could you close the issue?

PR #257, opened 2016-03-02
about ParamPattern and java.util.Date

ScienJus:
from https://github.com/decebals/pippo/issues/256

When pippo start, `DefaultControllerHandler.findMethod`will create some `ParameterValue` and call them `to` method, is there a better ways like call a method named `isSupport`? whatever, when method has a `java.util.Date` parameter, there will a error beacause `ParameterValue.to` is not support `java.util.Date` without `pattern`, and `DefaultControllerHandler.findMethod` do not parse `@ParamPatter`。

So I did these things:
1. when method has a `java.util.Date` parameter and without `@ParamPattern`, it will be converted  from a unix timestamp like `System.currentTimeMillis()`, because it's useful and not be supported by `@ParamPattern`.
2. in `DefaultControllerHandler.findMethod`, find all `@ParamPattern.value` from method parameters. and sent to `ParameterValue.to`

I could not find a good way to test it in unit test, but i create a real web application and it worked fine.

### DIFF
--- a/pippo-controller/src/test/java/ControllerHandlerTest.java
+++ b/pippo-controller/src/test/java/ControllerHandlerTest.java
@@ -19,26 +19,33 @@ import ro.pippo.controller.DefaultControllerHandler;
 import ro.pippo.core.Param;
 import ro.pippo.core.ParamPattern;
 
+import static org.junit.Assert.assertNotNull;
+
 /**
  * @author ScienJus
- * @date 2016/3/2.
  */
 public class ControllerHandlerTest {
 
     @Test
-    public void testFindMethodWithDate() throws Exception {
-        DefaultControllerHandler handler = new DefaultControllerHandler(DateController.class, "testDate");
-        handler = new DefaultControllerHandler(DateController.class, "testDateWithPattern");
+    public void testDateParam() throws Exception {
+        DefaultControllerHandler handler = new DefaultControllerHandler(DateController.class, "testDateParam");
+        assertNotNull(handler.getMethod());
+    }
+
+    @Test
+    public void testDateParamWithPattern() throws Exception {
+        DefaultControllerHandler handler = new DefaultControllerHandler(DateController.class, "testDateParamWithPattern");
+        assertNotNull(handler.getMethod());
     }
 
     static class DateController extends Controller {
 
-        public void testDate(@Param("date") java.util.Date since) {
+        public void testDateParam(@Param("date") java.util.Date since) {
             System.out.println(since);
         }
 
         // yyyy-MM-dd HH:mm is not supported from in java.sql.Date/Time/Timestamp, just for test
-        public void testDateWithPattern(@ParamPattern("yyyy-MM-dd HH:mm") @Param("date") java.util.Date since) {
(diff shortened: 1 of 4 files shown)


pippo-controller/src/test/java/ControllerHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+import org.junit.Test;
+import ro.pippo.controller.Controller;
+import ro.pippo.controller.DefaultControllerHandler;
+import ro.pippo.core.Param;
+import ro.pippo.core.ParamPattern;
+
+/**
+ * @author ScienJus
+ * @date 2016/3/2.
```
decebals: Please remove this line.

pippo-controller/src/test/java/ControllerHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class ControllerHandlerTest {
+
+    @Test
+    public void testFindMethodWithDate() throws Exception {
```
decebals: It's an idea to split this method in two methods (`testDateParam` and `testDateParamWithPattern`).
What we can do for both tests in the current implementation is to create a controller handler and to use an `assertNotNull(handler.getMethod())`.
I see a supplement assert related  to `parameterNames` if we create in `DefaultControllerHandler` a getter for `parameterNames` property. In this mode we can test the parameter names size and the paramater name.

decebals:
> when method has a java.util.Date parameter and without @ParamPattern, it will be converted from a unix timestamp like System.currentTimeMillis(), because it's useful and not be supported by @ParamPattern.

I don't know if this helps. I am not sure that I can supply you the unix timestamp from the frontend. In this situation maybe a hardcoded default date pattern is a better and simple idea.

@gitblit What you think?

Related to the code fragment from `ParameterValue`:
```java
if (pattern == null) {
    // no defined pattern, use type defaults
    if (type == java.sql.Date.class) {
        return toSqlDate();
    }

    if (type == Time.class) {
        return toSqlTime();
    }

    if (type == Timestamp.class) {
        return toSqlTimestamp();
    }
    // support java.util.Date
    if (Date.class.isAssignableFrom(type)) {
        //default from unix timestamp like System.currentTimeMillis()
        return toDateFromUnixTimestamp();
    }
} else {
    if (Date.class.isAssignableFrom(type)) {
        Date date = toDate(pattern);
        if (type == Date.class) {
            return date;
        } else if (type == java.sql.Date.class) {
            return new java.sql.Date(date.getTime());
        } else if (type == Time.class) {
            return new Time(date.getTime());
        } else if (type == Timestamp.class) {
            return new Timestamp(date.getTime());
        }
    }
}        
```

maybe it's better:
```java
if (Date.class.isAssignableFrom(type)) {
   // make something with 'pattern' if pattern is null // <-- HERE
    Date date = toDate(pattern);
    if (type == Date.class) {
        return date;
    } else if (type == java.sql.Date.class) {
        return new java.sql.Date(date.getTime());
    } else if (type == Time.class) {
        return new Time(date.getTime());
    } else if (type == Timestamp.class) {
        return new Timestamp(date.getTime());
    }
}
```

ScienJus:
ok, also people can modify the default date format is better.

Issue #232, opened 2015-12-17
Update to Pebble 2.1


gitblit:
Pebble updated to 2.1 and has breaking API changes.  It's probably a simple change, but I haven't investigated.

PR #261, opened 2016-04-04
Update to Pebble 2.2.1

gitblit:
Satisfies #232 .

There are some behavior differences to note in Pebble 2.  I'm still checking that in my apps but I think this update is solid.


decebals:
:+1:

Issue #262, opened 2016-04-14
Errors Configuring Underlying Tomcat Server to use HTTPS without Client Auth


JPercivall:
The problem arises when the keystore is properly set (location, password and alias) but no truststore is set. I get a file not found exception because it's looking for a file with name "null" (stack trace at end). 

Also of note, Tomcat isn't currently using the truststore because TomcatServer.java isn't set to use ClientAuth[1]. It would be nice to have this exposed as a TomcatSettings Property.

[1] https://github.com/decebals/pippo/blob/93f71655481f3133310019317d90467485f37e1f/pippo-tomcat/src/main/java/ro/pippo/tomcat/TomcatServer.java#L113-L113

SEVERE: Failed to load keystore type JKS with path null due to /Users/jpercivall/projects/edge/nifi-minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/target/tomcat_home/null (No such file or directory)
java.io.FileNotFoundException: /Users/jpercivall/projects/edge/nifi-minifi/minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/target/tomcat_home/null (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at java.io.FileInputStream.<init>(FileInputStream.java:93)
	at sun.net.www.protocol.file.FileURLConnection.connect(FileURLConnection.java:90)
	at sun.net.www.protocol.file.FileURLConnection.getInputStream(FileURLConnection.java:188)
	at org.apache.tomcat.util.file.ConfigFileLoader.getInputStream(ConfigFileLoader.java:74)
	at org.apache.tomcat.util.net.jsse.JSSESocketFactory.getStore(JSSESocketFactory.java:444)
	at org.apache.tomcat.util.net.jsse.JSSESocketFactory.getTrustStore(JSSESocketFactory.java:407)
	at org.apache.tomcat.util.net.jsse.JSSESocketFactory.getTrustManagers(JSSESocketFactory.java:658)
	at org.apache.tomcat.util.net.jsse.JSSESocketFactory.getTrustManagers(JSSESocketFactory.java:570)
	at org.apache.tomcat.util.net.NioEndpoint.bind(NioEndpoint.java:360)
[...]

PR #264, opened 2016-04-15
Use the trustStore only when client authentication is enabled

danjee:
Added the condition as @JPercivall suggested on issue #262


decebals:
Thanks!

JPercivall:
Just tested out this commit and it works as expected, thanks @danjee

PR #270, opened 2016-05-03
Make chunked transfer-encoding optional

gitblit:
By flushing the response buffer we force `chunked` transfer-encoding.  When we do this we lose the option to relay `Content-Length` to the client for `HEAD` requests or for setting-up client-side deterministic download progress indicators on `GET` requests.

This PR makes `chunked` transfer encoding optional and improves the `HEAD` response of the `DirectoryHandler`.


decebals:
Thanks!

Issue #273, opened 2016-05-17
Jetty and PippoTest


decebals:
I have a problem using Jetty, start the test and the test don't do anything, I believe that problem is caused by to the asm library.

For more info see https://groups.google.com/forum/#!topic/pippo-java/N-3GMVsXZW8

decebals:
I can reproduce this issue using a RouteTest in [pippo-demo-basic](https://github.com/decebals/pippo-demo/tree/master/pippo-demo-basic) module from pippo-demo project.
The problem is this line from JettyServer.start() method:
```java
server.join();
```
from [JettyServer]( https://github.com/decebals/pippo/blob/master/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java#L74)

If I comment this line the tests passed. My test class (test/java/... folder) looks like: 
```java
package ro.pippo.demo.basic;

import com.jayway.restassured.http.ContentType;
import com.jayway.restassured.response.Response;
import org.junit.Rule;
import org.junit.Test;
import ro.pippo.test.PippoRule;
import ro.pippo.test.PippoTest;

import static org.junit.Assert.assertEquals;

/**
 * @author Decebal Suiu
 */
public class RouteTest extends PippoTest {

    @Rule
    public PippoRule pippoRule = new PippoRule(new BasicApplication());

    @Test
    public void testIndex() {
        when().
            get("/").
        then().
            statusCode(200);
    }

    @Test
    public void testHello() {
        Response response = get("/");
        response.then().statusCode(200);
        response.then().contentType(ContentType.HTML);
        assertEquals("Hello World", response.asString());
    }

}
````

We must investigate if we can comment/delete line `server.join()` from the `JettyServer.start()` method.

barrantesgerman:
Hello

I investigated about `server.join();` from the `JettyServer.start()` and `tomcat.getServer().await();` from the `TomcatServer.start()`

barrantesgerman:
Hello

I investigated about `server.join();` from the `JettyServer.start()` and `tomcat.getServer().await();` from the `TomcatServer.start()`

In both cases the methods call [Thread.join()](http://docs.oracle.com/javase/1.5.0/docs/api/java/lang/Thread.html#join%28%29) and block the main thread of JUnit until the web server thread stops and release the main thread.

I searched how to make JUnit not work on the main thread, but I found nothing that would serve
[...]

PR #276, opened 2016-05-20
Workaround for Issue #273

barrantesgerman:
Hello

I investigated about `server.join();` from the `JettyServer.start()` and `tomcat.getServer().await();` from the `TomcatServer.start()`

In both cases the methods call [Thread.join()](http://docs.oracle.com/javase/1.5.0/docs/api/java/lang/Thread.html#join%28%29) and block the main thread of JUnit until the web server thread stops and release the main thread.

I searched how to make JUnit not work on the main thread, but I found nothing that would serve
I tried [this](http://www.codeaffine.com/2014/07/21/a-junit-rule-to-run-a-test-in-its-own-thread/) and [this](http://falutin.net/2012/12/30/multithreaded-testing-with-junit/)

We must not comment `server.join();` or `tomcat.getServer().await();` because as [AlexR](http://stackoverflow.com/questions/15924874/embedded-jetty-why-to-use-join) says:

> `join()` is blocking until server is ready. It behaves like Thread.join() and indeed calls join() of Jetty's thread pool. Everything works without this because jetty starts very quickly. However if your application is heavy enough the start might take some time. Call of `join()` guarantees that after it the server is indeed ready.

My final solution it's a workaround

**JettyServer.start()**
```java
if (RuntimeMode.getCurrent() != RuntimeMode.TEST) {
    server.join();
}
```
**TomcatServer.start()**
```java
if (RuntimeMode.getCurrent() != RuntimeMode.TEST) {
    tomcat.getServer().await();
}
```

According to many examples I saw on StackOverflow, in the case of JUnit you have to avoid `server.join();` and as [Joakim](http://stackoverflow.com/questions/19113359/sending-request-to-the-newly-created-jetty-server-for-testing-purposes) says is not needed for unit testing. 

But I'm no expert in JUnit, I do not know what consequences might have

### DIFF
--- a/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
+++ b/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
@@ -71,7 +71,7 @@ public class JettyServer extends AbstractWebServer<JettySettings> {
             String version = server.getClass().getPackage().getImplementationVersion();
             log.info("Starting Jetty Server {} on port {}", version, getSettings().getPort());
             server.start();
-            if (pippoSettings.isTest()) {
+            if (!pippoSettings.isTest()) {
                 server.join();
             }
         } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


decebals:
@barrantesgerman Congratulation for your explanation! It's very documented. I see a little modification.
I prefer to use `!pippoSettings.isTest()` instead of `RuntimeMode.getCurrent() != RuntimeMode.TEST`. Please make the modification and I will merge this PR.
Thanks!

decebals:
Thanks!

PR #283, opened 2016-06-20
Unit test

barrantesgerman:
Some unit tests for Memcached, Redis and MongoDB

### DIFF
--- a/pippo-session-parent/pippo-session-mongodb/src/test/java/ro/pippo/session/mongodb/MongoDBSessionDataStorageTest.java
+++ b/pippo-session-parent/pippo-session-mongodb/src/test/java/ro/pippo/session/mongodb/MongoDBSessionDataStorageTest.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2016 Herman Barrantes.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ro.pippo.session.mongodb;
+
+import com.mongodb.MongoClient;
+import de.flapdoodle.embed.mongo.MongodExecutable;
+import de.flapdoodle.embed.mongo.MongodProcess;
+import de.flapdoodle.embed.mongo.MongodStarter;
+import de.flapdoodle.embed.mongo.config.MongodConfigBuilder;
+import de.flapdoodle.embed.mongo.config.Net;
+import de.flapdoodle.embed.mongo.distribution.Version;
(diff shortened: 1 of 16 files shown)


pippo-session-parent/pippo-session-mongodb/src/test/java/ro/pippo/session/mongodb/MongoDBSessionDataStorageTest.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Copyright 2016 Herman Barrantes.
```
decebals: change this line please

pippo-session-parent/pippo-session-spymemcached/src/main/java/ro/pippo/session/spymemcached/SpymemcachedSessionDataStorage.java
```diff
@@ -69,7 +69,6 @@ public SessionData get(String sessionId) {
         if (obj == null) {
             return null;
         }
-        this.sessions.touch(sessionId, idleTime);
         SessionData sessionData = (SessionData) obj;
```
decebals: `return (SessionData) this.sessions.get(sessionId);`?

pippo-session-parent/pippo-session-mongodb/src/test/java/ro/pippo/session/mongodb/MongoDBSessionDataStorageTest.java
barrantesgerman: Done

pippo-session-parent/pippo-session-spymemcached/src/main/java/ro/pippo/session/spymemcached/SpymemcachedSessionDataStorage.java
barrantesgerman: Done

decebals:
Thanks!
In the future please don't start with `-` character in the title.
For example: simple "Change license" instead of "-Change license"

pippo-session-parent/pippo-session-spymemcached/src/main/java/ro/pippo/session/spymemcached/SpymemcachedSessionDataStorage.java
decebals: Just a curiosity. Why `return (SessionData) this.sessions.get(sessionId);` is not OK for you?
The actual code is:
```java
Object obj = this.sessions.get(sessionId);
return obj == null ? null : (SessionData) obj;
```
but in my opinion `obj == null ? null : (SessionData) obj;` is not needed, it's redundant.
It's only a curiosity, nothing wrong.

barrantesgerman: Because if the value is null, will give an error when doing the casting `(SessionData) null`

decebals: At compilation or at runtime?

barrantesgerman: My mistake, I thought I would get a ClassCastException, but not

PR #284, opened 2016-06-20
Pippo Metrics Parent

barrantesgerman:
This is the parent of all metrics modules

I replaced the tabs for 4 spaces in the poms


decebals:
:+1: Thanks!

barrantesgerman:
What would you like to be the next `parent`?

decebals:
Maybe `pippo-template-parent`.

PR #286, opened 2016-06-21
Isolating Hazelcast instance for each test

barrantesgerman:
I have made the suggested corrections

Any other recommendations is welcome

### DIFF
--- a/pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
+++ b/pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
@@ -17,11 +17,11 @@ package ro.pippo.session.hazelcast;
 
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
-import org.junit.AfterClass;
+import org.junit.After;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
-import org.junit.BeforeClass;
+import org.junit.Before;
 import org.junit.Test;
 import ro.pippo.session.SessionData;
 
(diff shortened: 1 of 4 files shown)


pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
```diff
@@ -33,19 +33,16 @@
...
     private static final String KEY = "KEY";
     private static final String VALUE = "VALUE";
-    private static HazelcastInstance hazelcastInstance;
-    private static HazelcastInstance hazelcastInstance2;
+    private HazelcastInstance hazelcastInstance;
 
-    @BeforeClass
-    public static void setUpClass() {
+    @Before
+    public void setUpClass() {
```
decebals: `setUp()` ?

pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
```diff
@@ -33,19 +33,16 @@
...
         hazelcastInstance = Hazelcast.newHazelcastInstance();
-        hazelcastInstance2 = Hazelcast.newHazelcastInstance();
     }
 
-    @AfterClass
-    public static void tearDownClass() {
+    @After
+    public void tearDownClass() {
```
decebals: `tearDown()` ?

barrantesgerman: I will change

decebals:
Test failed

barrantesgerman:
Yes, I will check

barrantesgerman:
I think I found the problem

Use `put` instead of `set` to guarantee the replication in the cluster
Use `remove` instead of `delete` to guarantee the replication in the cluster
Set 2 async backups to guarantee the copy in the unit test

For more information
http://docs.hazelcast.org/docs/3.6/manual/html-single/index.html#backing-up-maps

decebals:
I see that you have fun doing these implementations :smile:  I am sure that you learn a lot with this occasion.

barrantesgerman:
Yes, it is part of what I like about this job, I am forced to learn something new every day

barrantesgerman:
What would you like to be the next `parent`?

decebals:
The next `parent` can be `pippo-content-type-parent`. I prefer to work with short names, something similar with `pippo-abc-parent` but I feel that it's not possible right now (I don't know if we can change content-type with a simple word).

barrantesgerman:
I think the same, but right now can not think of another name

PR #287, opened 2016-06-21
Pippo Content Type Parent

barrantesgerman:
This is the parent of all content type modules


decebals:
Super. Thanks!

PR #288, opened 2016-06-21
Extract addRoute, GET, POST, ... methods in Routing interface

decebals:
The idea is that I can add routes directly in `Pippo` class without use `getApplication()` method.

I don't know if `Routing` is a good/clear name. Or maybe `RouteAware` is better. My English is not so good so I cannot help in this direction.

What you say? Is it a valuable PR?


gitblit:
Yes, I like this.  :+1: for using default methods!

decebals:
@gitblit Thanks James for feedback. You saved this PR :smile:

PR #289, opened 2016-06-24
Pippo Server Parent

barrantesgerman:
This is the parent of all server modules


decebals:
Thanks! :+1:

barrantesgerman:
What would you like to go next?

I think it could be `pippo-controller-parent`

decebals:
Sure. I think that we must add `pippo-spring`, `pippo-guice` and `pippo-weld` in controller-parent because  these modules have sense only for controllers.

PR #290, opened 2016-06-24
Pippo Controller Parent

barrantesgerman:
This is the parent of all controller modules


decebals:
Super. Thanks!

Issue #293, opened 2016-06-28
JettyServer.start() blocks due to Jetty server.join()


sl33nyc:
This is related to #273.

Using v0.8, I want to start up some background threads that depend on the web server being fully started, but the call to JettyServer.start() also blocks the main thread from initializing any web server dependencies.  The Spark micro web service framework handles this using a [`CountDownLatch`](https://github.com/perwendel/spark/blob/master/src/main/java/spark/embeddedserver/jetty/EmbeddedJettyServer.java#L129).

In master, I see that the runtime environment will affect whether `server.join()` is called, but then an equivalent, `public` method for `.join()` needs to be implemented across all web server implementations.

decebals:
I have a solution. I will resolve today the issue.

decebals:
Can you help me with a minimum unit test?

decebals:
@sl33nyc, @barrantesgerman Please try the fix. I updated the SNAPSHOT.
I will make some tests to see if I must to do the same thing for Tomcat. I know that Undertow is not affected by this behavior.

decebals:
I tested all servers (jetty, tomcat, undertow and tjws). Everything is OK. I am impressed to see that the tiny TJWS makes a good job :smile:

sl33nyc:
👍

barrantesgerman:
Sorry, this week I've been kind of busy but I'll try the fix

barrantesgerman:
Sorry for taking so long, I've been kind of busy.

I did some tests and I think that all is well.

I would like to share the test is performed if help someone else.

https://github.com/barrantesgerman/pippo-test-server

I discovered two interesting things about Rest Assured

* If you want to test using the session, you must add:
```java
SessionFilter sessionFilter = new SessionFilter();
given().filter(sessionFilter)
```
* If you want to test using any of the pippo-session modules, you should also add:
```java
RestAssured.config = RestAssured.config().sessionConfig(new SessionConfig().sessionIdName("SESSIONID"));
```
To use CookieSessionStrategy

decebals:
@barrantesgerman Thanks

PR #294, opened 2016-06-29
Verify that starting server does not block

sl33nyc:
Suggested tests for #293.

I didn't add the tests for the other backend types, but ideally, all `WebServer` implementations should have the same behavior.


decebals:
I tested with https://github.com/decebals/pippo/issues/273#issuecomment-219800990

PR #302, opened 2016-07-27
New EncryptorSessionDataTranscoder

barrantesgerman:
New implementation of `SessionDataTranscoder` to encrypt the data.

```java
@Override
protected RequestResponseFactory createRequestResponseFactory() {
    SessionDataStorage sessionDataStorage = new CookieSessionDataStorage(getPippoSettings(), new EncryptorSessionDataTranscoder("PasswordPassword"));
    SessionManager sessionManager = new SessionManager(sessionDataStorage);

    return new SessionRequestResponseFactory(this, sessionManager);
}
```


decebals:
I don't think that we need a new module. `pippo-session` is the perfect place to put the new `EncryptedSessionDataTranscoder`.
Allso, in the future please try to make few commits for a PR (for example in this PR is enough a single commit) because it's more clear when we look in the (git) log.
Also, in the new `EncryptedSessionDataTranscoder` (and in general) I prefer to not use telescopic constructors. Maybe fluent setter methods (that returns this) or a inner Builder class (build pattern) is better.

barrantesgerman:
Sorry for the delay :worried: , I make the requested changes :smile: 

```java
    @Override
    protected RequestResponseFactory createRequestResponseFactory() {
        EncryptorSessionDataTranscoder transcoder = EncryptorSessionDataTranscoder.builder().secretKey("PasswordPassword").build();
        SessionDataStorage sessionDataStorage = new CookieSessionDataStorage(getPippoSettings(), transcoder);
        SessionManager sessionManager = new SessionManager(sessionDataStorage);

        return new SessionRequestResponseFactory(this, sessionManager);
    }
```

As always, any suggestions are welcome

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/EncryptorSessionDataTranscoder.java
```diff
@@ -0,0 +1,128 @@
...
+    private final SessionDataTranscoder transcoder;
+
+    public EncryptorSessionDataTranscoder(String secretKey, String hmacSHA1Key, SessionDataTranscoder transcoder, Encryptor encryptor) {
+        this.secretKey = secretKey;
+        this.hmacSHA1Key = hmacSHA1Key;
+        this.transcoder = transcoder;
+        this.encryptor = encryptor;
+    }
+
+    public static Builder builder() {
```
decebals: We don't need this method EncryptedSessionDataTranscoder.Builder is enough.

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/EncryptorSessionDataTranscoder.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public static class Builder {
+
+        private String secretKey;
+        private String hmacSHA1Key;
+        private Encryptor encryptor;
+        private SessionDataTranscoder transcoder;
+
+        Builder() {
```
decebals: Remove this constructor. The code is:
```java
new EncryptedSessionDataTranscoder.Builder.secretKey(secretKey).build();
```

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/EncryptorSessionDataTranscoder.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+package ro.pippo.session;
+
+import ro.pippo.core.PippoRuntimeException;
+import ro.pippo.core.util.CryptoUtils;
+
+/**
+ * @author Herman Barrantes
+ */
+public class EncryptorSessionDataTranscoder implements SessionDataTranscoder {
```
decebals: I think that `EncryptedSessionDataTranscoder` is a better name.

barrantesgerman:
Applied all suggested changes :+1:

decebals:
Thanks!

PR #303, opened 2016-08-03
Reduce DirectoryHandler logging noise for normal IOExceptions

gitblit:
Normal behaviors like `connection reset by peer`, `broken pipe`, and `connection timed out` **with** their stacktrace clog the logs with useless information.  Instead log the IOException message, if available, with metadata about the request.


decebals:
Thanks!

PR #311, opened 2016-09-12
Fix #310

### LINKED ISSUE #310: MongoDBSessionDataStorageTest is failing
While running test in my local, when the port 27017 is already running, the test fails. This because test runs on the hardcoded port which will lead for the failure.

anantheshadiga:
While running test in my local, when the port 27017 is already running, the test fails. This because test runs on the hardcoded port which will lead for the failure

### DIFF
--- a/pippo-session-parent/pippo-session-mongodb/src/test/java/ro/pippo/session/mongodb/MongoDBSessionDataStorageTest.java
+++ b/pippo-session-parent/pippo-session-mongodb/src/test/java/ro/pippo/session/mongodb/MongoDBSessionDataStorageTest.java
@@ -39,21 +39,21 @@ public class MongoDBSessionDataStorageTest {
     private static final String VALUE = "VALUE";
     private static final String DATABASE_NAME = "embedded";
     private static final String HOST = "localhost";
-    private static final Integer PORT = 27017;
     private static MongodExecutable mongodExe;
     private static MongodProcess mongod;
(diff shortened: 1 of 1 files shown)


pippo-session-parent/pippo-session-mongodb/src/test/java/ro/pippo/session/mongodb/MongoDBSessionDataStorageTest.java
```diff
@@ -50,10 +49,10 @@ public static void setUpClass() throws IOException {
         mongodExe = starter.prepare(
                 new MongodConfigBuilder()
                 .version(Version.Main.PRODUCTION)
-                .net(new Net(PORT, Network.localhostIsIPv6()))
+                .net(new Net(Network.getFreeServerPort(), Network.localhostIsIPv6()))
                 .build());
         mongod = mongodExe.start();
-        mongoClient = new MongoClient(HOST, PORT);
+        mongoClient = new MongoClient(HOST, Network.getFreeServerPort());
```
decebals: I don't know if the second call of `Network.getFreeServerPort()` returns the same port (another variant is to use an internal variable `port = Network.getFreeServerPort()`). 
If you are sure that the actual code is OK I will merge it.

anantheshadiga:
@decebals corrected it

decebals:
Thanks!

Issue #320, opened 2016-10-20
Register servlet/filter


decebals:
https://groups.google.com/forum/#!topic/pippo-java/kdDMJEHN6Bg

decebals:
Closed by PR #321

PR #321, opened 2016-10-20
Work #320

decebals:
How to add a servlet (DemoServlet):

```java
public class DemoServlet extends HttpServlet {

    @Override
    protected void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
        PrintWriter out = response.getWriter();
        out.println("<html>");
        out.println("<body>");
        out.println("<h1>Demo Servlet</h1>");
        out.println("</body>");
        out.println("</html>");
    }

}

public class ServletApplication extends Application {

    @Override
    protected void onInit() {
        ServletContext servletContext = getServletContext();
        // add servlet 
        ServletRegistration.Dynamic demoServlet = servletContext.addServlet("demo", DemoServlet.class);
        demoServlet.setLoadOnStartup(1);
        demoServlet.addMapping("/demo");
        // other possible settings for demoServlet

        // add regular routes 
        GET("/", routeContext -> routeContext.send("Hello from Pippo route!"));
    }

}

public class ServletDemo {

        Pippo pippo = new Pippo(new ServletApplication());
        pippo.getServer().setPippoFilterPath("/app/*");
        pippo.start();
    }

}
```

So, the effective code is in `ServletApplication.onInit()` where you must retrieve the `ServletContext` instance (via new method `getServletContext()`) and call `addServlet()` or `addFilter()`.
That it's all. 
The modifications to accommodate this feature are minimal and "safe" in my opinion.

Open your browser and type:
* http://localhost:8338/app (for Pippo route)
* http://localhost:8338/demo (for Demo servlet)

I tested with Jetty, Tomcat and Undertow. 
On Jetty and Tomcat everything is OK but on Undertow I encountered a problem (an error). I see that the error in Undertow is generated by the code from `ServletApplication.onInit()` where I add the new servlet.

@gitblit Can you take a look to see if you can resolve the problem with Undertow?

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/AbstractWebServer.java
+++ b/pippo-core/src/main/java/ro/pippo/core/AbstractWebServer.java
@@ -15,16 +15,26 @@
  */
 package ro.pippo.core;
 
+import javax.servlet.ServletContextListener;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
 /**
  * @author Decebal Suiu
  */
-public abstract class AbstractWebServer<T extends WebServerSettings> implements WebServer {
+public abstract class AbstractWebServer<T extends WebServerSettings> implements WebServer<T> {
 
     protected PippoFilter pippoFilter;
     protected String pippoFilterPath;
 
     protected PippoSettings pippoSettings;
     private T settings;
+
+    protected List<Class<? extends ServletContextListener>> listeners;
+
+    public AbstractWebServer() {
+        listeners = new CopyOnWriteArrayList<>();
+    }
 
     protected abstract T createDefaultSettings();
 
@@ -43,7 +53,7 @@ public abstract class AbstractWebServer<T extends WebServerSettings> implements 
     }
 
     @Override
-    public WebServer setPippoFilter(PippoFilter pippoFilter) {
+    public WebServer<T> setPippoFilter(PippoFilter pippoFilter) {
         this.pippoFilter = pippoFilter;
 
         return this;
@@ -55,7 +65,7 @@ public abstract class AbstractWebServer<T extends WebServerSettings> implements 
     }
 
     @Override
-    public WebServer setPippoFilterPath(String pippoFilterPath) {
+    public WebServer<T> setPippoFilterPath(String pippoFilterPath) {
         PippoFilter.validateFilterUrlPattern(pippoFilterPath);
 
         this.pippoFilterPath = pippoFilterPath;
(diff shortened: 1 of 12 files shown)


decebals:
In demo we must change the default Pippo filter path from `/*` to `/app/*` because otherwise the filter "eat" all requests ("/demo" in our demo)

decebals:
The error in Undertow is:
```
Context Path:
Servlet Path:/app
Path Info:null
Query String:null
Stack Trace
javax.servlet.ServletException: java.lang.IllegalStateException: UT010041: The servlet context has already been initialized, you can only call this method from a ServletContainerInitializer or a ServletContextListener
ro.pippo.core.PippoFilter.init(PippoFilter.java:107)
io.undertow.servlet.core.LifecyleInterceptorInvocation.proceed(LifecyleInterceptorInvocation.java:111)
```

decebals:
You cannot add servlet, filter or listener if the context is in initialize phase (`PippoFilter,init() -> Application.init() -> Application.onInit()`), so forget about the solution with `ServletApplication.onInit()` if you want to use Undertow (it works only for `Jetty` and `Tomcat` because the implementation of the servlet specifications is not so restricted).
The universal solution (tested on all major servers - Jetty, Tomcat, Undertow) is to add the servlet in a `WebServerInitializer`.

```java
import org.kohsuke.MetaInfServices;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import ro.pippo.core.WebServerInitializer;

import javax.servlet.ServletContext;
import javax.servlet.ServletRegistration;

@MetaInfServices
public class ServletAppender implements WebServerInitializer {

    private static final Logger log = LoggerFactory.getLogger(ServletAppender.class);

    @Override
    public void init(ServletContext servletContext) {
        ServletRegistration.Dynamic demoServlet = servletContext.addServlet("demo", DemoServlet.class);
        demoServlet.setLoadOnStartup(1);
        demoServlet.addMapping("/demo");
        // other possible settings for demoServlet

        log.debug("Added servlet '{}' to '{}'", demoServlet.getClassName(), demoServlet.getMappings().iterator().next());
    }

    @Override
    public void destroy(ServletContext servletContext) {
        // do nothing
    }

}
```

```java
mport ro.pippo.core.Pippo;

import static ro.pippo.core.route.Route.GET;

public class ServletDemo {

        Pippo pippo = new Pippo();

        // set pippo filter path
        pippo.getServer().setPippoFilterPath("/app/*");

        // add route
        GET("/", routeContext -> routeContext.send("Hello from Pippo route!"));

        pippo.start();
    }

}
```

I am OK with the current solution. In the future I can add a new method `WebServer.addInitializer(WebServerInitializer)` for people that want to bypass the ServiceLoader mechanism.

decebals:
Added [pippo-demo-servlet](https://github.com/decebals/pippo-demo/tree/master/pippo-demo-servlet) that demonstrate how you can use a servlet and a filter via `WebServerInitializer` and `WebServer.addListener`.

PR #323, opened 2016-10-31
Inject Application instance in ServletContext

decebals:
In last time I tried to improve the interoperability between `Application` and `WebServer`.
For now we can add `Filter`, `Servlet` and `Listener` in a generic/uniform mode for all builtin `WebServer` types via `WebServerInitializer` and `WebServer.addListtener(ServletContextListener listener)`.
Practically we can customize in a easy mode many aspects of the `WebServer`. Using `Pippo.setServer(WebServer server)` we can customize any aspect of the embedded server.
With the current PR I propose you a simple but powerful solution to improve the interoperability between `Application` and other services (for example __Jersey__, __Guice__ via GuiceServletContextListener and others).
With this PR you have access to the `Application` instance from `ServletConetxt`.

Below, I will show you a practical case, how I can integrate Jersey in Pippo with with very few lines of code.

```java

import javax.ws.rs.GET;
import javax.ws.rs.Path;

@Path("/")
public class MyResource {

    @GET
    @Path("/hello")
    public String getMessage() {
        return "Hello World";
    }

}

import org.glassfish.jersey.server.ResourceConfig;

public class MyApplication extends Application {

    public ResourceConfig getResourceConfig() {
        return new ResourceConfig(MyResource.class);
     }
     
    // other possible methods

}

import ro.pippo.core.WebServer;

@MetaInfServices
public class JerseyInitializer implements WebServerInitializer {

    @Override
    public void init(ServletContext servletContext) {
        // get the resourceConfig via application
        MyApplication application = (MyApplication) servletContext.getAttribute(PIPPO_APPLICATION);
        ResourceConfig resourceConfig = application.getResourceConfig();

        // add jersey servlet and mapping this servlet on "/api/*"
        ServletRegistration.Dynamic jerseyServlet = servletContext.addServlet("jersey", new ServletContainer(resourceConfig));
        jerseyServlet.setLoadOnStartup(1);
        jerseyServlet.addMapping("/api/*");
    }

    @Override
    public void destroy(ServletContext servletContext) {
        // do nothing for now
    }

}

public class JerseyDemo {

        Pippo pippo = new Pippo(new MyApplication());

        // set pippo filter path
        pippo.getServer().setPippoFilterPath("/app/*");

        pippo.start();
    }

}
``` 

In above example you can see that `JerseyInitializer` can be extracted easily in a `pippo-jersey` module (probably with an abstract method `createResourceConfig():ResourecConfig` or a simple convention `Application.getLocal("JERSEY_RESOURCE_CONFIG"):ResourceConfig`). 

I think this is the beginning of a beautiful friendship between `WebServer` (container and components) and Pippo :smile:

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/WebServer.java
+++ b/pippo-core/src/main/java/ro/pippo/core/WebServer.java
@@ -22,6 +22,55 @@ import javax.servlet.ServletContextListener;
  */
 public interface WebServer<T extends WebServerSettings> {
 
+    /**
+     * Attribute name used to retrieve the application instance from a {@link javax.servlet.ServletContext}.
+     * See {@link WebServerInitializer} and {@link ServletContextListener}.
+     *
+     * <pre>
+     * {@code
+     * MyApplication application = (MyApplication) servletContext.getAttribute(PIPPO_APPLICATION);
+     * }
+     * </pre>
+     *
+     * A possible scenario: I want to add support for Jersey in my application.
+     * <pre>
+     * {@code
+     *
+     * public class MyApplication extends Application {
+     *
+     *     public ResourceConfig getResourceConfig() {
+     *         return new ResourceConfig(MyResource.class);
+     *     }
+     *
+     *     // other possible methods
+     *
+     * }
+     *
+     * @MetaInfServices
+     * public class JerseyInitializer implements WebServerInitializer {
+     *
+     *     @Override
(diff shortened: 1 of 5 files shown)


gitblit:
Excellent improvement. Makes Pippo even more flexible than ever.

decebals:
@gitblit Thanks my friend!

Issue #331, opened 2016-12-06
Handling Accept-Encoding request


munendrasn:
How to handle `Accept-Encoding` header in request?
Suppose `Accept-Encoding=gzip` then, response header should contain `Content-Encoding=gzip`.
How to achieve this?
Just adding header to response wouldn't be sufficient. Is there way already available in `pippo` to handle this or custom implementation is required for this??

decebals:
> Just adding header to response wouldn't be sufficient.

Can you give me more details? Can you supply a code snippet to show me what you try to achieve?

decebals:
You can wrap the response's output stream in a `GzipOutputStream` and write to it.

```java
response.contentType(...);
GzipOutputStream outputStream = new GzipOutputStream(response.getOutputStream());
// write to outputStream
```

For some servlet containers (like Tomcat) you can obtain this feature automatically using configuration (a parameter).

If my code snippet is OK for you, I can add a new method `setGzip` (or something similar) in `Response` class that set automatically the content type and wrap the output stream in a `GzipOutputStream`.

munendrasn:
I think creating a method in `Response` class would be better, so that this can be automatically set

decebals:
Can you confirm that the above snippet resolves your problem?

decebals:
My idea here is to supply an out of the box solution, something that is transparent for end user/developer and that could be implemented via [HttpServletResponseWrapper](http://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponseWrapper.html) or by introduction of a new `ResponseWrapper` concept.

munendrasn:
Apologizes for the delayed response. The above snippet solved my problem.
Instead of introducing new `ResponseWrapper`, this can be handled in `Response` class right??
To handle it in `Response` class,  access to `request` is needed.

munendrasn:
I think your idea is introduce `ResponseWrapper` something like in this [link](http://tutorials.jenkov.com/java-servlets/gzip-servlet-filter.html) right?

decebals:
Yes, something like that.

The idea is to inject `GzipOutputStream` directly in `HttpServletRequest` (via `HttpServletRequestWrapper.getOutputStream` method) if the request contains `gzip` in `Accept-Encoding`, to set `gzip` in `Content-Encoding` in HttpServletResponse` and to do nothing Pippo's Response (gzip is transparent). So all responses will be "gzipped" out of the box.

munendrasn:
[...]

PR #333, opened 2016-12-13
Use gzip compression where it's possible; see #331

decebals:
See #31 for more details.

I made some tests and everything is OK. 
So, this compression is used out of the box (you as developer do nothing) if the client (the browser - I tested on Chrome) accept "gzip" as encoding.
I made a little refactoring, `RequestResponseFactory` contains only a method `createRequestResponse(HttpServletRequest, HttpServletResponse):RequestResponse`. Practically I merged  `createRequest` and `createResponse` methods in one method `createRequestResponse` because are multiple situation when I need to have access from response to request and vice versa. I hope that this modification doesn't affect you.

Of course any advice, question is welcome.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/Application.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Application.java
@@ -17,6 +17,7 @@ package ro.pippo.core;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import ro.pippo.core.gzip.GZipRequestResponseFactory;
 import ro.pippo.core.route.DefaultRouter;
 import ro.pippo.core.route.ResourceRouting;
 import ro.pippo.core.route.Route;
@@ -248,7 +249,8 @@ public class Application implements ResourceRouting {
      * @return
      */
     protected RequestResponseFactory createRequestResponseFactory() {
-        return new RequestResponseFactory(this);
+//        return new RequestResponseFactory(this);
+        return new GZipRequestResponseFactory(this);
     }
 
     /**
--- a/pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
+++ b/pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
@@ -21,7 +21,6 @@ import org.junit.After;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
-import org.junit.Before;
 import org.junit.Test;
 import ro.pippo.session.SessionData;
 
@@ -33,16 +32,10 @@ public class HazelcastSessionDataStorageTest {
     private static final String SESSION_NAME = "session";
     private static final String KEY = "KEY";
     private static final String VALUE = "VALUE";
-    private HazelcastInstance hazelcastInstance;
-
-    @Before
-    public void setUp() {
-        hazelcastInstance = Hazelcast.newHazelcastInstance();
-    }
 
     @After
     public void tearDown() {
-        hazelcastInstance.shutdown();
+        Hazelcast.shutdownAll();
     }
 
     /**
@@ -51,7 +44,7 @@ public class HazelcastSessionDataStorageTest {
     @Test
     public void testCreate() {
         System.out.println("create");
-        HazelcastSessionDataStorage instance = new HazelcastSessionDataStorage(hazelcastInstance);
+        HazelcastSessionDataStorage instance = new HazelcastSessionDataStorage(Hazelcast.newHazelcastInstance());
         SessionData sessionData = instance.create();
         sessionData.put(KEY, VALUE);
         assertNotNull(sessionData);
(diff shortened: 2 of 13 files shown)


pippo-core/src/main/java/ro/pippo/core/Application.java
```diff
@@ -248,7 +249,8 @@ public final RequestResponseFactory getRequestResponseFactory() {
      * @return
      */
     protected RequestResponseFactory createRequestResponseFactory() {
-        return new RequestResponseFactory(this);
+//        return new RequestResponseFactory(this);
+        return new GZipRequestResponseFactory(this);
```
decebals: I don't know if it's a good idea or not to have `GZipRequestResponseFactory` (gzip encoding) as default (instead of simple `RequestResponseFactory`). What do you say?

pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
```diff
@@ -65,18 +65,19 @@ public void testCreate() {
      */
     @Test
     public void testSave() {
-        HazelcastInstance hazelcastInstance2 = Hazelcast.newHazelcastInstance();
+//        HazelcastInstance hazelcastInstance2 = Hazelcast.newHazelcastInstance();
```
decebals: @barrantesgerman I remarked that if I don't comment the second instance of Hazelcast, I obtain a test fail (`saved` is null). I don't understand why?

munendrasn: @decebals I think this may be due this Hazelcast [issue](https://github.com/hazelcast/hazelcast/issues/8871)

decebals: It's not a blocker for me. I don't understand why we need a second instance in that test (I am a newbie on Hazelcast).

pippo-core/src/main/java/ro/pippo/core/Application.java
munendrasn: @decebals , instead of creating separate `RequestResponseFactory`(like `GZipRequestResponseFactory`), the base class itself could be used for this purpose right?
Because what if `Accept-Encoding` specifies `deflate` instead of `gzip`. Also, it would be easier to extend for other new features by using common `RequestResponseFactory`

decebals: Maybe there are people that don't want this feature builtin. From this reason I created `GZipRequestResponseFactory` that extends the simple `RequestResponseFactory`.
Also I want to mention that you can extends `GZipRequestResponseFactory` and to override `acceptsGZipEncoding` if you want more control (only some urls, ...).
What I don't like me is that I cannot use (yet) the wrapper/decorator pattern for `RequestResponseFactory`. For example if I want gzip compression on the fly I should use `GZipRequestResponseFactory`, if I want a different `SessionStorage` (for example Hazelcast) I should use `SessionResquestResponseFactory`. Now it's not possible to use both `GZipRequestResponseFactory` and `SessionResquestResponseFactory`.

pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
barrantesgerman: The second instance is just to test how it works with a distributed session.

I am not an expert on Hazelcast, but I will investigate the problem in search of a solution.

decebals: @barrantesgerman Thanks. I don't know why `testSave` on this PR (in my opinion is nothing wrong in this). Please see first if that test fail on master branch.

pippo-core/src/main/java/ro/pippo/core/Application.java
barrantesgerman: I think like @decebals , and hold `RequestResponseFactory` as it is.

I also consider that in the future you could create a wrapper or decorator to wrap several future implementations of `RequestResponseFactory`

pippo-session-parent/pippo-session-hazelcast/src/test/java/ro/pippo/session/hazelcast/HazelcastSessionDataStorageTest.java
barrantesgerman: My apologies, I don't know how to do a checkout of a pull request, I will search it on Google, but if you want give me some advice, you are welcome.

decebals: @barrantesgerman `git checkput -b gzip origin/gzip`
A PR is a possible merge from a branch (`gzip` in this case).

anantheshadiga:
@decebals I know you are probably looking at the code coverage and code quality during merging of any code. Lets not allow any redundant commit like the one which we have in this PR https://github.com/decebals/pippo/pull/333/commits/b900bd219a77a2dac12130ce79fa787e1a18a425. Any  PR's should be rebased properly and the commits should be legitimate.

I would even appreciate we follow the commit message guidelines, like the [angular.js](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines) has one, unless it mentions the fixing of github issue.

decebals:
> Any PR's should be rebased properly and the commits should be legitimate.

You are right. I will try to respect this rule in the future when I will accept a PR.

> I would even appreciate we follow the commit message guidelines, like the angular.js has one, unless it mentions the fixing of github issue.

I will try to take a look. If the guidelines for the commit messages are not complicated we can use them.

Thanks for your observations!

uknth:
@decebals TBH, any sort of standardisation is a good idea. It makes the project much more readable and will help you in review process. (angular.js is just one of the many standards) 
We i.e. me and @anantheshadiga are more than happy to help. We are actively using pippo in production at our organisation @unbxd and would love to contribute more in future.

decebals:
@uknth Thanks! I am open to any proposals (coming from anybody) that bring value to this project.

PR #336, opened 2016-12-16
add getWriter method to Response

### LINKED ISSUE #335: Wrong content-type in the response
```java
routeContext.setHeader("content-type", "application/xml") // this wont work
routeContext.getResponse().getHttpServletResponse().getWriter()
//write something
```
In this case, response content-type would be `text/html`. It is because, in `Response` class, `getHeaderMap()` stores all added headers and adds to `httpServletResponse` only when `finalizeResponse()` is called.
In the above case, `finalizeResponse` method is never called.

Possible solutions:

- expose a `getWriter()` method in `Response` class to fetch `writer` 
```java
getWriter() { 
try{ 
htttpServeltResponse.getWriter()
} catch (IOException e) {
 // throw pippo runtime
}
}
```
-  instead of adding new header to `getHeaderMap` add directly to `httpServletResponse`

Let me know if there are any other better ways. Please let me know, If any of these approach is good  I will raise a PR for the same

munendrasn:
@decebals Please review
Fixes #335

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/Response.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Response.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.StringWriter;
+import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


munendrasn:
Test cases are failing because of this [issue](https://github.com/flapdoodle-oss/de.flapdoodle.embed.memcached/issues/11)

pippo-core/src/main/java/ro/pippo/core/Response.java
```diff
@@ -26,13 +26,7 @@
...
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletResponse;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.StringWriter;
+import java.io.*;
```
decebals: I prefer to not use wildcard import. Can you please fix this line? If you use IntelliJ Idea you can take a look at [this](http://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports).
Thanks

munendrasn: I have removed wildcard import

decebals:
Thanks!

PR #339, opened 2017-01-04
Fixes #338

### LINKED ISSUE #338: Consistency in Parameter Value conversion
For following url, `example.com?q=&format=json`

while converting q to different data types, response would be as follows
```
routeContext.getParameter("q").toInt() # gives 0(default value)
routeContext.getParameter("q").toString() # gives empty string
routeContext.getParameter("q").toList() # gives a list containing empty string
```
Shouldn't `toList()` return empty list??

This is because in `toList` method in `ParameterValue.java`,  only null check is done.
It would be better to check for this condition also

```
if (isNull() || (values.length==1 && StringUtils.isNullorEmpty(values[0])))
```
This would return `defaultValue` (in this case, empty list).  or is there a reason for present approach??
If my approach looks good, I will raise a PR for this one

munendrasn:
Return empty list when `toList()` is called for parameter with empty string as its value

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/ParameterValue.java
+++ b/pippo-core/src/main/java/ro/pippo/core/ParameterValue.java
@@ -239,7 +239,7 @@ public class ParameterValue implements Serializable {
     }
 
     public Set<String> toSet(Set<String> defaultValue) {
-        if (isNull()) {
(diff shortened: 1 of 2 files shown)


decebals:
Thanks!

PR #346, opened 2017-01-27
feat(Response): expose method to reset the response

munendrasn:
@decebals,
This is required because on calling `httpServletResponse.reset()`, only response gets reset. But other params like headers, cookies also need to erased from `Response` class

### DIFF
--- a/pippo-core/src/test/java/ro/pippo/core/ResponseTest.java
+++ b/pippo-core/src/test/java/ro/pippo/core/ResponseTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ro.pippo.core;
+
+import org.junit.Test;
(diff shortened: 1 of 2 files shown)


pippo-core/src/test/java/ro/pippo/core/ResponseTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+
+/**
+ * @author munendrasn
+ */
+@RunWith(MockitoJUnitRunner.class)
```
decebals: I don't think that we need this annotation. This annotation is useful if you use mock annotation, for example
```java
@Mock
private HttpServletResponse servletResponse;
```

instead of
```java
private HttpServletResponse servletResponse = mock(HttpServletResponse.class);
```

pippo-core/src/test/java/ro/pippo/core/ResponseTest.java
```diff
@@ -0,0 +1,62 @@
...
+package ro.pippo.core;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import javax.servlet.http.HttpServletResponse;
+
+import java.nio.charset.StandardCharsets;
+
```
decebals: extra empty line (please use optimize import from IntelliJ for example)

decebals:
I am happy to see unit test code.
Also, as a curiosity, when you use `Response.rest()` method (a use case). Thanks!

munendrasn:
@decebals ,
In some cases, I'm writing directly to `writer` or `outputStream`. Also, setting headers for response using `response.header()` or `routeContext.setHeader()`.
Now, If any error or exception occurs in application, all set headers need to be removed  (using `servletResponse.reset()`).

But, this doesn't reset headers, cookies in `Response`. Hence, exposing this helper will help in required reset

Issue #324, opened 2016-11-01
Stateless singleton controller to handle requests


mkhanal:
Issue:
Stateful controllers + server states with sessions  = Stateful servers

Details Issue: 
Controller module is very handy as it takes away the boilerplate needed to parse parameters.

Following the layered architecture, if layers(controller/services/...) are singleton and stateless, then layers do not need to be instantiated for every route and thus never need to be garbage collected. This makes app very fast, and at the same time helps push the stateless-ness forward. Since in pippo, unlike rails, the Controllers do not need to be stateful to bind values to its templates, we can just as easily make controllers stateless.

While using controller module
A typical route:
```java
handlerClass = SomeController.class;
GET(path, handlerClass, "methodName");
```
Internally, this keeps the reference of the class and evaluates method so that on every request, the class is re-instantiated and the method is invoked. 

Proposal: 
Accept accept an instance instead of class as : 
```java
handlerInstance= new SomeController();
GET(path, handlerInstance, "methodName");
```
or even
```java
handlerInstance = new SomeController();
GET(path, handlerInstance::methodName)
```
Now this instance can be reused for every call. This also resembles what the application would look like if we do not use the ControllerApplication, and use the ro.pippo.core.Application 's route handling.

decebals:
Thanks for the detail explanation (excellent description of problem).
I don't know yet what approach is better:
- add the instance as singleton (route)
- add the class and the method name (controller)
- add the method reference (controller)
- add a kind of "metadata" - for example class name as string (route and controller)

All the variants have strengths and weaknesses.
For example, if I add the instance or class (use an import java statement in Application with a concrete class name) I'm in trouble when I try to implement the Hot Deployment feature [1]. If I use one `RouteHandler` for all requests (the standard route implementation) someone reports a problem [2] when try to execute requests in parallel. In [3] you will get some details about the current implementation.
In the feature my intention is to use a super declarative approach [4] for pippo-controller, similar with SpringBoot. Definitel we must evolve the `Controller` concept and the whole `pippo-controller` module.

In the meantime I will think to the solution proposed by you. The problem is the backward compatibility. Maybe we find an elegant and simple solution for all cases. 

Thanks again!

[1] https://groups.google.com/d/msg/pippo-java/lTbVWX3ArRs/lrr6XJ9KCQAJ
[2] https://groups.google.com/forum/#!topic/pippo-java/vZaa-0GxU1U
[3] https://github.com/decebals/pippo/issues/263
[...]

PR #350, opened 2017-02-07
Add possibility to use a singleton Controller

decebals:
This PR offers you the possibility to specify a Controller instance for all requests (singleton).

```java
addControllers(ContactsController.class); // one instance for EACH request
// or
addControllers(new ContactsController()); // one instance for ALL requests
```

See #324 for details.


decebals:
The `Controller` class plays only a decorative role, it contains only helper method:
```java
public class Controller {

    public final RouteContext getRouteContext() {
        return RouteDispatcher.getRouteContext();
    }

    public final Request getRequest() {
        return getRouteContext().getRequest();
    }

    public final Response getResponse() {
        return getRouteContext().getResponse();
    }

    @SuppressWarnings("unchecked")
    public <T extends ControllerApplication> T getApplication() {
        return (T) getRouteContext().getApplication();
    }

}
``` 

In theory, in the future, we can remove the actual constrain that a controller must extends the `Controller` class.

Issue #356, opened 2017-03-29
some bug


lambdaupb:
https://github.com/decebals/pippo/blob/f143f3172933d35c50265e039323ff8dc3a03dce/pippo-core/src/main/java/ro/pippo/core/route/ClasspathResourceHandler.java#L37

```
    String resourceName = getResourceBasePath() + "/" + resourcePath;
    return this.getClass().getClassLoader().getResource(resourceName);
```

Missing path normalization leads to a path traversal vulnerability.

I tested it in my own app and was able to retrieve `META-INF/` resource files via a manual GET request with `/.../META-INF/` inserted.

This is really web security 101 and maybe the other handler should be audited for similar vulnerabilities.

lambdaupb:
My apologies, I should have sent this via a private channel so this could be published with a fixed version ready to use.

But it seems impossible to edit the GitHub issue in a way that hides the information needed to abuse it. (https://github.com/isaacs/github/issues/37)
I have changed the title to something boring though.

decebals:
Thanks for reporting the problem! I will investigate the issue as much as possible.

decebals:
Ups, I committed on other branch (reloading) but now it's OK because I merged that branch on master. It's tricky when you work on multiple branches/features :smile:

PR #357, opened 2017-03-29
#356 test and partial fix

lambdaupb:
This hopefully fixes the `ClasspathResourceHandler` (and with it `PublicResourceHandler`) while the `WebjarsResourceHandler` is still insecure.

A test is implemented that checks some of the resource handlers with a naive case of this issue.

see #356


pippo-core/src/main/java/ro/pippo/core/route/ClasspathResourceHandler.java
```diff
@@ -15,31 +15,41 @@
...
 
-        this.resourceBasePath = getNormalizedPath(resourceBasePath);
+        this.resourceBasePath = Paths.get(getNormalizedPath(resourceBasePath));
     }
 
     @Override
     public URL getResourceUrl(String resourcePath) {
-        String resourceName = getResourceBasePath() + "/" + resourcePath;
-        return this.getClass().getClassLoader().getResource(resourceName);
+        Path resolved = resourceBasePath.resolve(resourcePath).normalize();
```
decebals: I prefer to keep `resourceBasePath` as String and to modify only `getResourceUrl`.
The modification create a `Path requestedPath` variable and check to see if this path startWith resourceBasePath (similar with https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/route/DirectoryHandler.java#L138)

pippo-core/src/test/java/ro/pippo/core/PathTraversalResourceHandlerTest.java
```diff
@@ -0,0 +1,53 @@
+package ro.pippo.core;
```
decebals: Please add copyright header (see other .java files).

pippo-core/src/test/java/ro/pippo/core/PathTraversalResourceHandlerTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class PathTraversalResourceHandlerTest {
+
+  @Test
```
decebals: Something is wrong with formatting. We use tab or 4 characters for indentation.

pippo-core/src/main/java/ro/pippo/core/route/ClasspathResourceHandler.java
```diff
@@ -32,10 +38,20 @@ public ClasspathResourceHandler(String urlPath, String resourceBasePath) {
...
+    private static Path ROOT_PATH = Paths.get("/");
+    private static String normalizePath(String subPath) {
+        return ROOT_PATH.relativize(ROOT_PATH.resolve(subPath).normalize()).toString();
+    }
+
     @Override
     public URL getResourceUrl(String resourcePath) {
-        String resourceName = getResourceBasePath() + "/" + resourcePath;
-        return this.getClass().getClassLoader().getResource(resourceName);
+        String normalizedPath = normalizePath(resourcePath);
```
decebals: Is it not much simple to have a method:
```
// we can choose other name, if you feel that actual name is not suggestive
boolean validateResourcePath(String root, String path) {
    // put the logic here
}
```

and use this method in both `ClasspathResourceHandler` and `WebjarsResourceHandler`?

lambdaupb: Sure, but I wont touch the `WebjarsResourceHandler`, I don't understand how it should work and don't really have the time to dive in. Feel free to change anything or not use this pull request.

decebals:
Fixed with last commit.
@lambdaupb I appreciate very much your contribution. Thanks!

PR #360, opened 2017-04-20
Websocket support

decebals:
It's the successor of #279. All information presented in #279 are available.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/AbstractWebServer.java
+++ b/pippo-core/src/main/java/ro/pippo/core/AbstractWebServer.java
@@ -24,10 +24,10 @@ import java.util.concurrent.CopyOnWriteArrayList;
  */
 public abstract class AbstractWebServer<T extends WebServerSettings> implements WebServer<T> {
 
-    protected PippoFilter pippoFilter;
+    private PippoFilter pippoFilter;
     protected String pippoFilterPath;
 
-    protected PippoSettings pippoSettings;
+    private Application application;
     private T settings;
 
     protected List<Class<? extends EventListener>> listeners;
@@ -49,12 +49,25 @@ public abstract class AbstractWebServer<T extends WebServerSettings> implements 
 
     @Override
     public PippoFilter getPippoFilter() {
+        if (pippoFilter == null) {
+            setPippoFilter(createPippoFilter());
+        }
+
         return pippoFilter;
     }
 
+    /**
+     * Set the {@link PippoFilter} instance.
+     * This method call {@link PippoFilter#setApplication(Application)} to end.
+     *
+     * @param pippoFilter
+     * @return
+     */
     @Override
     public WebServer<T> setPippoFilter(PippoFilter pippoFilter) {
         this.pippoFilter = pippoFilter;
+
+        pippoFilter.setApplication(application);
 
         return this;
     }
@@ -74,8 +87,8 @@ public abstract class AbstractWebServer<T extends WebServerSettings> implements 
     }
 
     @Override
-    public WebServer<T> init(PippoSettings pippoSettings) {
-        this.pippoSettings = pippoSettings;
+    public WebServer<T> init(Application application) {
+        this.application = application;
 
         return this;
     }
@@ -85,6 +98,28 @@ public abstract class AbstractWebServer<T extends WebServerSettings> implements 
         listeners.add(listener);
 
         return this;
+    }
+
+    public Application getApplication() {
+        return application;
+    }
+
+    /**
+     * Override this method if you want to customize the {@link PippoFilter}.
+     * <p/>
+     * <pre>
+     * protected PippoFilter createPippoFilter() {
+     *     PippoFilter pippoFilter = super.createPippoFilter();
+     *     pippoFilter.setIgnorePaths(Collections.singleton("/favicon.ico"));
+     *
+     *     return pippoFilter;
+     * }
+     * </pre>
+     *
+     * @return
+     */
+    protected PippoFilter createPippoFilter() {
+        return new PippoFilter();
     }
 
 }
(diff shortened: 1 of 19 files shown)


decebals:
> - [ ] inject WebSocketFilter automatically in Pippo launcher 

See https://github.com/decebals/pippo/pull/279#issuecomment-224386236 for more details.
Resolved in a smooth way by 7c4c069.

mhagnumdw:
Hello, Decebals!

The dependency _org.eclipse.jetty.websocket:websocket-server_ is optional after your commit https://github.com/decebals/pippo/commit/7c4c069caed9b02757eeaa20d9641c146ffcc18b. How to tell maven to add this dependency so that JettyServer.java add the JettyWebSocketFilter filter instead of PippoFilter?

decebals:
@mhagnumdw 
It's made automatically by `JettyServer` in `createPippoFilter` [method](https://github.com/decebals/pippo/commit/7c4c069caed9b02757eeaa20d9641c146ffcc18b#diff-31482071eb48761b0a733c58d84c3f5fR197).
All you have to do is just add Jetty's `websocket-server` as dependency in your `pom.xml`:
```
<!-- Enable websocket support -->
<dependency>
    <groupId>org.eclipse.jetty.websocket</groupId>
    <artifactId>websocket-server</artifactId>
    <version>${jetty.version}</version>
</dependency>
```
The advantages are:
- Jetty server is minimal for non websocket application, as in present
- `JettyServer` class from Pippo adds automatically `JettyWebSocketFilter` instead of `PippoFilter` (via `createPippoFilter` method) if it "feels" that you added in your classpath (via pom.xml) a dependency to websocket implementation from Jetty

A new `pippo-demo-websocket` module is prepared already and it will be available in `pippo-demo` project.   

You can see parts from this demo in [this](https://gist.github.com/decebals/4b16c78ab4879ece3324ea79d21bc7c2) Gist.

The websocket support is implemented only for Jetty in this moment.
The implementation is finished, the API is stable. I will add a little bit more documentation in this PR and after this we can remove a new Pippo version that contains the websocket support.

mhagnumdw:
@decebals 

Thanks for the answer. I had already done so. I just find it a bit strange because the developer can specify a version for the _org.eclipse.jetty.websocket: websocket-server_ that conflicts with the version of the Jetty server shipped in Pippo, right? Is there a way for the developer not to specify the version so that it is obtained from the Jetty server version used by Pippo? I went crazy? :)

Other inquiries,
1) Is it possible to send a message to a group? From what I've seen (and already tested!) It's only possible to send a message to a single user and to all (brodcast);
2) Is it possible to send a message from an asynchronous process?

PS: Is it better to address these issues here or in the mailing list?

decebals:
They are some situations when you want to access some details information in your `WebSocketHandler`. For example, I want to find the remote address, the request URI, the HTTP method or the parameter map.
For this situation we have a very simple solution (see below the code):
```java
addWebSocket("/ws-debug", (webSocketContext, message) -> {
    WebSocketConnection connection = webSocketContext.getConnection();
    // cast WebSocketConnection to JettyWebSocketConnection
    JettyWebSocketConnection jettyConnection = (JettyWebSocketConnection) connection;
    String remoteAddress = jettyConnection.getRemoteAddress().toString();
    System.out.println("remoteAddress = " + remoteAddress);
    org.eclipse.jetty.websocket.api.Session session = jettyConnection.getSession();
    String protocolVersion = session.getProtocolVersion();
    System.out.println("protocolVersion = " + protocolVersion);
    org.eclipse.jetty.websocket.api.UpgradeRequest upgradeRequest = session.getUpgradeRequest();
//    upgradeRequest.getCookies();
//    upgradeRequest.getHeaders();
//    upgradeRequest.getParameterMap();
    String query = upgradeRequest.getQueryString();
    System.out.println("query = " + query);
});
```

Now, if I make a request from a websocket client (I used `Simple WebSocket Client` extension for Chrome) with the url `ws://localhost:8338/ws-debugt?param1=a&param2=b`, I obtain the below result in console:

```
remoteAddress = /127.0.0.1:35594
protocolVersion = 13
query = test=true
```  

That it's all, simple and straightforward :smile:

decebals:
@mhagnumdw 
I am not a Maven expert, so I don't know if it's possible to not specify the jetty version in your pom.xml. I putted in my demo the same version as in `pippo-jetty` module. I think that it's important/mandatory to have the same version. 
I feel that my approach with websocket-server as optional is correct, because the majority of applications will be non websocket and I don't want to add extra dependency, but it's a possibility that I may be wrong.

> Is it possible to send a message to a group? From what I've seen (and already tested!) It's only possible to send a message to a single user and to all (brodcast)

I think that you can. `WebSocketContext` contains all clients connections, the current client connection and two helper methods for sending a message (to current client or to all clients - broadcast). You can create a method that filters your clients according some criteria (the group) and to send the message only to these clients. I will expose the `connections` field from `WebSocketContext` via `getConnections` method to have access to all connections for that websocket.

> Is it possible to send a message from an asynchronous process?

I don't know what involves this. Can you add more details about your use case?

decebals:
@mhagnumdw 
> Is it possible to send a message to a group? From what I've seen (and already tested!) It's only possible to send a message to a single user and to all (brodcast)

Other solution is to use the websocket hook methods `onOpen` and `onClose` from `WebSocketHandler` and to keep a map in your Application with all connections grouped on a criteria. In `onOpen` method you add a new entry and `onClose` method you delete an existed entry. If you have this map it's super easy to send a message for all clients from a group.

mhagnumdw:
@decebals 

> > Is it possible to send a message from an asynchronous process?
> 
> I don't know what involves this. Can you add more details about your use case?
> 

I have a periodic schedule made by _java.util.concurrent.ScheduledExecutorService_. This asynchronous task collects some data that application users should be aware of. To achieve this, I imagined this asynchronous task to post the messages and get them to the client via websocket (so the client / user does not have to keep giving F5 on the screen for news!).

decebals:
@mhagnumdw 
I don't see a problem. The approach described by me in comment https://github.com/decebals/pippo/pull/360#issuecomment-296467537 is a good start point.

mhagnumdw:
@decebals 
Thank you very much! I'll be working on proofs of concept in the coming days.

Issue #274, opened 2016-05-17
customize jackson objectMapper


amalagaura:
I wanted to use JacksonMixInAnnotations to customize the serialization of JSON objects which I cannot control via annotations. It means I need access to the objectMapper instance.

Objectmapper is available via the getObjectMapper method implemented by JacksonJsonEngine.

If I want to override it, do I need to duplicate the JacksonBaseEngine code from pippo-jackson and then add another service initializer? I think I cannot include the pippo-jackson jar because it will cause JacksonJsonEngine to be loaded with the default objectMapper?

Thank you,

decebals:
@gitblit Can you take a look please?

gitblit:
@amalagaura I would...

1. include the `pippo-jackson` dependency
2. subclass `JacksonJsonEngine` and override `getObjectMapper()`
3. call `application.getContentTypeEngines().setContentTypeEngine(new CustomJacksonEngine())`;

amalagaura:
Thank you. That was my first thought as well, but I was concerned it would register multiple contentTypeEngines because of the initializers. 

But I think you are saying that the pippo-jackson would get initialized and register the default but my code would be in my application therefore since setContentTypeEngine writes to a map it would just get overwritten. (sorry, thinking out loud I guess).

thank you

ab2971082:
ok, thanks!

PR #376, opened 2017-07-21
Support configuring the Jackson Object Mapper

lorrin:
This change provides a getter for the Jackson Object Mapper so that Pippo applications can customize it as they need. E.g. I have this in my `onInit`:
```Java
ContentTypeEngine applicationJsonEngine = getContentTypeEngine(APPLICATION_JSON);
if (applicationJsonEngine instanceof JacksonJsonEngine) {
    JacksonJsonEngine jacksonJsonEngine = (JacksonJsonEngine) applicationJsonEngine;
    ObjectMapper objectMapper = jacksonJsonEngine.getObjectMapper();
    objectMapper.registerModule(new Jdk8Module());
    objectMapper.registerModule(new GuavaModule());
} else {
    throw new IllegalStateException(String.format(
            "Cannot launch without Jackson engine registered to handle %s content.", APPLICATION_JSON));
}
```

### DIFF
--- a/pippo-content-type-parent/pippo-jackson/src/main/java/ro/pippo/jackson/JacksonBaseEngine.java
+++ b/pippo-content-type-parent/pippo-jackson/src/main/java/ro/pippo/jackson/JacksonBaseEngine.java
@@ -39,17 +39,13 @@ public abstract class JacksonBaseEngine implements ContentTypeEngine {
 
     @Override
     public void init(Application application) {
-        objectMapper = constructObjectMapper();
+        objectMapper = createObjectMapper();
         objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
         objectMapper.setTimeZone(TimeZone.getDefault());
         objectMapper.registerModule(new AfterburnerModule());
     }
 
-    public ObjectMapper getObjectMapper() {
-        return objectMapper;
-    }
-
-    protected abstract ObjectMapper constructObjectMapper();
+    protected abstract ObjectMapper createObjectMapper();
 
     @Override
     public String toString(Object object) {
--- a/pippo-content-type-parent/pippo-jackson/src/main/java/ro/pippo/jackson/JacksonBaseEngine.java
+++ b/pippo-content-type-parent/pippo-jackson/src/main/java/ro/pippo/jackson/JacksonBaseEngine.java
@@ -23,7 +23,6 @@ import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.module.afterburner.AfterburnerModule;
 import ro.pippo.core.Application;
 import ro.pippo.core.ContentTypeEngine;
-import ro.pippo.core.HttpConstants;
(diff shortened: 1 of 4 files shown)


decebals:
Thanks! I will take a look.
It's possible to customize Jackson Object Mapper, see https://github.com/decebals/pippo/issues/274.

decebals:
In the end, I don't see the the difference. You replaced `getObjectMapper` method with `constructObjectMapper` (by the way, I believe that `createObjectMapper` is a better name for a factory method) .
Correct me if I am wrong?

pippo-content-type-parent/pippo-jackson/src/main/java/ro/pippo/jackson/JacksonBaseEngine.java
```diff
@@ -36,17 +35,21 @@
...
     @Override
     public void init(Application application) {
-        objectMapper = getObjectMapper();
+        objectMapper = constructObjectMapper();
         objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
         objectMapper.setTimeZone(TimeZone.getDefault());
         objectMapper.registerModule(new AfterburnerModule());
     }
 
-    protected abstract ObjectMapper getObjectMapper();
```
decebals: It's enough to make this method public.

lorrin:
Hi, I hadn't looked at #274 before submitting this. I agree `create` instead of `construct` would be a better name for the factory method.

It is not enough to make `getObjectMapper` public, because it creates a new ObjectMapper every time it is called. I wanted to be able to get the already-created ObjectMapper and make my own modifications to it.

I think the #274 pattern is probably cleanest, even though a bit more verbose. I propose the following: 
1) rename the existing creation method to `create`
2) add a `get` that returns the previously created OM
3) leave `get` protected so that users are pushed to follow the #274 pattern

What would be different is that when I create my own ObjectMapper, I would just extend `JacksonJsonEngine` and override it's create method to do super, get the result, apply the changes. I don't have to care about the abstract parent class or reimplementing things it does that I want to preserve. I'll update this PR.

lorrin:
Never mind, I hadn't fully understood the #274 pattern until I tried it. You're right, the changes I submitted weren't needed to customize Jackson easily. I've kept this PR alive as a simple rename from `get` to `create` for the factory method. Thanks for the help.

decebals:
The rename is good but we have a problem because we brake the API, without a very good reason and I am sure that they are people that use this method (`getObjectMapper`) already.

lorrin:
Alright, I'll close this then. Thanks for your help.

lorrin:
BTW, if anyone stumbles across this looking for info on how to follow the #274 pattern, it's like this:
```java
        JacksonJsonEngine customEngine = new JacksonJsonEngine(){
                @Override
                public void init(Application application) {
                    super.init(application);
                    objectMapper.registerModule(new Jdk8Module());
                    objectMapper.registerModule(new GuavaModule());
                }
            };
        customEngine.init(this);
        getContentTypeEngines().setContentTypeEngine(customEngine);
```

PR #377, opened 2017-07-21
Handle keystore paths that are relative to Classpath

lorrin:
This might need some further work, but I thought I'd share to get the discussion started. I wanted to put a dummy self-signed cert in my JAR to use when in dev mode, and I discovered that I couldn't figure out a way to set `server.keystoreFile` in `application.properties` that would load it. If running from a JAR, Jetty wants a jar:file:/... style path. But if just running in my IDE, I need a file-system relative path. With this change I can add `%dev.server.keystoreFile = /pippo_self_signed.pkcs12` to my application.properties and it works in both cases without further modification. (`pippo_self_signed.pkcs12` resides in my `src/main/resources` folder)

### DIFF
--- a/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
+++ b/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
@@ -40,6 +40,11 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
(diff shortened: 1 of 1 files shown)


decebals:
I need to understand your use case.
You have self-signed cert and you want to use this cert only in `dev` mode. You run your application two modes:
- via (fat) jar, the cert is included in jar file
- via IDE, the cert is in classpath (src/main/resources)

lorrin:
Yes, that is correct.

More broadly speaking, I'm interested in making sure I've got Pippo set up properly for TLS while I'm developing, and this seemed like the right way to do it. If you think there's a better way, please let me know. :-)

lorrin:
Hi @decebals , is there anything I can do to help move this PR forward?

decebals:
Thanks!

Issue #375, opened 2017-07-19
Jetty startup issues are silently dropped


lorrin:
When Jetty fails to start, the exception is caught and wrapped in a PippoRuntimeException (https://github.com/decebals/pippo/blob/master/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java#L126) but then this exception silently vanishes inside the Executor that invoked `internalStart`. https://github.com/decebals/pippo/blob/master/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java#L75 waits forever for `internalStart` to succeed.

I'm a bit confused by the purpose of the Executor here. If `start` is going to block until the executor completes, why doesn't it just do the startup logic on its own thread?

As a work-around, I've added some logging when this happens.
```
$ git diff
diff --git a/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java b/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
index cd96706d..753d307e 100644
--- a/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
+++ b/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
@@ -123,6 +123,7 @@ public class JettyServer extends AbstractWebServer<JettySettings> {
                 server.join();
             }
         } catch (Exception e) {
+            log.error("Unable to launch Jetty", e);
             throw new PippoRuntimeException(e);
         }
     }
```

Common examples are port already in use, keystore configuration issues.

decebals:
Very good observation. In last period I have been trying to improve readability for `*Server` classes.
If you take a look to the [history](https://github.com/decebals/pippo/commits/master/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java) of `JettyServer` you will see that the executor appears when I resolved https://github.com/decebals/pippo/issues/293.
The idea is that the line [server.join()](https://github.com/decebals/pippo/blob/master/pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java#L123) block the thread, and this action is not good if you want to run some code after Jetty/Pippo start (see https://github.com/decebals/pippo/issues/293, https://github.com/decebals/pippo/issues/273 and https://github.com/decebals/pippo/issues/371).
[...]

PR #378, opened 2017-07-21
Jetty startup

lorrin:
This is the simplified Jetty Startup I proposed to address #375

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/Pippo.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Pippo.java
@@ -163,7 +163,7 @@ public class Pippo implements ResourceRouting, ReloadWatcher.Listener {
 
     public void stop() {
         if (!running) {
-            log.warn("Server is not started");
+            log.debug("Server is not started");
             return;
         }
 
@@ -245,10 +245,7 @@ public class Pippo implements ResourceRouting, ReloadWatcher.Listener {
     }
 
     private void addShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-            if (running)
-                stop();
-        }));
+        Runtime.getRuntime().addShutdownHook(new Thread(Pippo.this::stop));
     }
 
     private Application createApplication() {
--- a/pippo-core/src/main/java/ro/pippo/core/Pippo.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Pippo.java
@@ -245,7 +245,10 @@ public class Pippo implements ResourceRouting, ReloadWatcher.Listener {
     }
 
     private void addShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(new Thread(Pippo.this::stop));
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            if (running)
+                stop();
+        }));
     }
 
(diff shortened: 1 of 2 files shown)


pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
```diff
@@ -58,46 +55,8 @@
 
     private Server server;
 
-    private final ExecutorService executor;
-    private final CountDownLatch  startLatch;
-
-    public JettyServer() {
```
decebals: I think that we need this constructor with `super()`. See https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/AbstractWebServer.java#L36.

pippo-core/src/main/java/ro/pippo/core/Pippo.java
```diff
@@ -245,7 +245,10 @@ public void onEvent(ReloadWatcher.Event event, Path dir, Path path) {
     }
 
     private void addShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(new Thread(Pippo.this::stop));
```
decebals: I think that we should preserve the original line (more compact) because the check of `running` flag is done in `stop()`.

pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
lorrin: The noarg superconstructor should be called by default: https://stackoverflow.com/a/2054040

pippo-core/src/main/java/ro/pippo/core/Pippo.java
lorrin: I agree about stop. My thinking was that stop logs a warning when you call it twice, so I should be sure not to call it twice. But maybe the better change is to change the log to `debug` since calling stop twice is normal?

lorrin:
OK, I've adjusted the `stop()` method.

BTW, the threading doesn't seem totally safe here. `running` and `reloading` are `volatile`, but there is unsynchronized logic that checks the current value, then does work and  updates the value. Two threads could call `start`, both see `running==false`, then both proceed to try to start the server. Instead of `synchronized` blocks, it might be better to fix this with an AtomicReference to an Enum that represents the possible states. Then if you want to start the server, you see if it is in Uninitialized state. If it is, try to compareAndSet it to Starting state. if you fail, then probably some other thread was already starting it. If you succeed, you get to start it. Then when you're done, set it to Started state. Etc. But that's unrelated to this PR. Also only a hypothetical concern and not a real-world problem for me so far. ;-)

pippo-server-parent/pippo-jetty/src/main/java/ro/pippo/jetty/JettyServer.java
lorrin: I put some logging (not checked in) to confirm the AbstractWebServer constructor is indeed still being called.

decebals:
Thanks!

lorrin:
Sure, happy to help. Thanks for being so response to all the PRs.

Issue #359, opened 2017-04-13
Set Undertow max Parameter


munendrasn:
Using `pippo-undertow` in my project. I want to increase default `MAX_PARAMETER` limit.

Only way to achieve this is to extend `UndertowServer` and setting `serverOptions` in createServer.
Is there a better way to do this??

Probably like adding in `application.properties` and picking the value using `pippoSettings`.
Then we could set all serverOptions in `createServer()` itself

decebals:
Your approach is good.
Is it a possibility to use `application.properties` to define options and use `setServerOption` (or `setSocketOption`, `setWorkerOption`) from `Undertow.Builder` class.
We can use a prefix in application.properties, like:
```
undertow.server.<option_name_1>=...
undertow.server.<option_name_2>=...
undertow.socket.<option_name_3>=...
undertow.socket.<option_name_4>=...
undertow.worker.<option_name_5>=...
undertow.worker.<option_name_6>=...
```

In `UndertowSettings` you can add `getServerOptions` (`getSocketOptions`, `getWorkerOptions`) method|(s) that get all settings with prefix `undertow.server` (`undertow.socket`, `undertow.worker`) using [PippoSettings.getSettingNames(String startingWith)](https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/PippoSettings.java#L418), remove the prefix from each returned setting and use http://docs.jboss.org/xnio/3.0.4.GA/api/org/xnio/Option.html#fromString(java.lang.String,java.lang.ClassLoader) to convert a string option in an Option instance.

It's just a suggestion, I am not sure if it works.

munendrasn:
I will try this approach and revert back to you with results

munendrasn:
> In UndertowSettings you can add getServerOptions (getSocketOptions, getWorkerOptions) method|(s) that get all settings with prefix undertow.server (undertow.socket, undertow.worker) using PippoSettings.getSettingNames(String startingWith), remove the prefix from each returned setting and use http://docs.jboss.org/xnio/3.0.4.GA/api/org/xnio/Option.html#fromString(java.lang.String,java.lang.ClassLoader) to convert a string option in an Option instance.

Before trying out this approach, a small question. We would get `List<String>` from `PippoSettings.getSettings()` and convert them to `Options`. But we wont be knowing about the `option` type right??
Same would be true for Option value(both would be same)

For Example,
Consider this `application.properties` file
```
[...]

PR #384, opened 2017-08-25
Add support to configure undertow

munendrasn:
- Pick undertow options from pippoSettings.
- All parameters would be in this format undertow.<type>.<parameter>
type has 3 values - server, socket, worker

Currently maintaining map in `UndetowSettings` which maintains mapping between parameter and its type.
If this approach is okay, I will add more parameters to map in `UndertowSettings` to the parameter list as extensive. 
Also, Could you suggest way to add any parameter without using map (currently map has to be maintained to know parameter type)
This fix is related to issue #359

### DIFF
--- a/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowServer.java
+++ b/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowServer.java
@@ -147,6 +147,8 @@ public class UndertowServer extends AbstractWebServer<UndertowSettings> {
             }
         }
 
+        // add undertow options
+        getSettings().addUndertowOptions(builder);
         builder.setHandler(contextHandler);
 
         return builder.build();
--- a/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowSettings.java
+++ b/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowSettings.java
@@ -17,37 +17,31 @@ package ro.pippo.undertow;
 
 import io.undertow.Undertow;
 import io.undertow.UndertowOptions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.xnio.Option;
-import org.xnio.OptionMap;
 import ro.pippo.core.PippoSettings;
 import ro.pippo.core.WebServerSettings;
 
-import java.util.HashMap;
+import java.lang.reflect.Field;
 import java.util.List;
-import java.util.Map;
 
 /**
  * @author Decebal Suiu
  */
 public class UndertowSettings extends WebServerSettings {
 
+    private static final Logger log = LoggerFactory.getLogger(UndertowServer.class);
+
     public static final String BUFFER_SIZE = "undertow.bufferSize";
+    private static final String PREFIX = "undertow.";
+    private static final String UNDERTOW_SERVER_PREFIX = "undertow.server.";
+    private static final String UNDERTOW_WORKER_PREFIX = "undertow.worker.";
+    private static final String UNDERTOW_SOCKET_PREFIX = "undertow.socket.";
     public static final String BUFFERS_PER_REGION = "undertow.buffersPerRegion";
     public static final String DIRECT_BUFFERS = "undertow.directBuffers";
     public static final String IO_THREADS = "undertow.ioThreads";
     public static final String WORKER_THREADS = "undertow.workerThreads";
-    private static Map<String, Class> parameterOptionMap;
-
-    private interface UndertowParameters {
-        String SERVER_PREFIX = "undertow.server.";
-        String WORKER_PREFIX = "undertow.worker.";
-        String SOCKET_PREFIX = "undertow.socket.";
-
-        // add all undertow parameters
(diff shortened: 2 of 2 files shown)


decebals:
I love to see contributions.

> Could you suggest way to add any parameter without using map (currently map has to be maintained to know parameter type)
My idea is to use reflection on Options.class (UndertowOptions.class) to extract all options (static final fields by Option<?> type).

munendrasn:
Even if we get all options by this approach, we still have to store their type because
when adding server options using `builder.setServerOptions()` we should know the type of option and passing `Option<?>` would not work . In that case we need to cast it (again mapping between option and type).

decebals:
I don't pass `Option<?>`.
The workflow is:
- read `undertow.*`settings via PippoSettings
- for each undertow setting retrieve the Option via reflection on UndertowOptions.class; retrieve the value for the static field that have the name equals with undertow settings (for example if I have `undertow.server.MAX_HEADER_SIZE` I retrieve via reflection the `UndertowOptions.MAX_HEADER_SIZE` value that is an `Option<Integer>`).

The problem is if for an object Option<T> (Option<Integer> in our case) we can extract the generic type and I feel that is possible.

decebals:
Maybe with something like this:
```java
String className = ((ParameterizedType) myOption.getClass().getGenericSuperclass()).getActualTypeArguments()[0].getTypeName();
```

decebals:
Maybe with something like this:
```java
String className = ((ParameterizedType) myOption.getClass().getGenericSuperclass()).getActualTypeArguments()[0].getTypeName();
```

decebals:
See below the code (maybe it can be improved):
```java
package com.mycompany;

import io.undertow.UndertowOptions;
import org.xnio.Option;

import java.lang.reflect.Field;

public class OptionTest {

        Field[] declaredFields = UndertowOptions.class.getDeclaredFields();
        for (Field field : declaredFields) {
            String fieldName = field.getName();
            if (Option.class.getName().equals(field.getType().getTypeName())) {
                try {
                    Object value = field.get(null);
                    Field f = value.getClass().getDeclaredField("type");
                    f.setAccessible(true);
                    // extract class name from a string like "class java.lang.Integer"
                    String typeClassName = f.get(value).toString().substring(6);
                    System.out.println(fieldName + " > " + typeClassName);
                } catch (IllegalAccessException | NoSuchFieldException e) {
                    e.printStackTrace();
                }
            }
        }
    }

}
```

The output is:
```
MAX_HEADER_SIZE > java.lang.Integer
MAX_ENTITY_SIZE > java.lang.Long
MULTIPART_MAX_ENTITY_SIZE > java.lang.Long
BUFFER_PIPELINED_DATA > java.lang.Boolean
IDLE_TIMEOUT > java.lang.Integer
REQUEST_PARSE_TIMEOUT > java.lang.Integer
NO_REQUEST_TIMEOUT > java.lang.Integer
MAX_PARAMETERS > java.lang.Integer
MAX_HEADERS > java.lang.Integer
MAX_COOKIES > java.lang.Integer
ALLOW_ENCODED_SLASH > java.lang.Boolean
DECODE_URL > java.lang.Boolean
URL_CHARSET > java.lang.String
ALWAYS_SET_KEEP_ALIVE > java.lang.Boolean
ALWAYS_SET_DATE > java.lang.Boolean
MAX_BUFFERED_REQUEST_SIZE > java.lang.Integer
RECORD_REQUEST_START_TIME > java.lang.Boolean
ALLOW_EQUALS_IN_COOKIE_VALUE > java.lang.Boolean
ENABLE_SPDY > java.lang.Boolean
ENABLE_HTTP2 > java.lang.Boolean
ENABLE_STATISTICS > java.lang.Boolean
ENABLE_CONNECTOR_STATISTICS > java.lang.Boolean
ALLOW_UNKNOWN_PROTOCOLS > java.lang.Boolean
HTTP2_SETTINGS_HEADER_TABLE_SIZE > java.lang.Integer
HTTP2_SETTINGS_ENABLE_PUSH > java.lang.Boolean
HTTP2_SETTINGS_MAX_CONCURRENT_STREAMS > java.lang.Integer
HTTP2_SETTINGS_INITIAL_WINDOW_SIZE > java.lang.Integer
HTTP2_SETTINGS_MAX_FRAME_SIZE > java.lang.Integer
HTTP2_SETTINGS_MAX_HEADER_LIST_SIZE > java.lang.Integer
HTTP2_PADDING_SIZE > java.lang.Integer
HTTP2_HUFFMAN_CACHE_SIZE > java.lang.Integer
MAX_CONCURRENT_REQUESTS_PER_CONNECTION > java.lang.Integer
MAX_QUEUED_READ_BUFFERS > java.lang.Integer
MAX_AJP_PACKET_SIZE > java.lang.Integer
REQUIRE_HOST_HTTP11 > java.lang.Boolean
MAX_CACHED_HEADER_SIZE > java.lang.Integer
HTTP_HEADERS_CACHE_SIZE > java.lang.Integer
SSL_USER_CIPHER_SUITES_ORDER > java.lang.Boolean
```

munendrasn:
@decebals I have used approach suggested by you. Using this all `UndertowOptions` would be supported. Let me know if any changes required for the PR

decebals:
Thanks! Good job.

PR #388, opened 2017-09-15
TemplateEngine customization and extension support 

codematix:
This change allows customizing the file extension used by `PebbleTemplateEngine` from the default of `.peb`. The customization can be applied by adding the following line in `conf/application.properties`:

```properties
template.extension = pebble
```

The same can be used by other pippo template integrations.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
+++ b/pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright (C) 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ro.pippo.core;
+
+import ro.pippo.core.route.RouteContext;
+import ro.pippo.core.route.Router;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
+
+import java.io.Writer;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Convenience abstract implementation of {@link TemplateEngine}
+ *
+ * A configuration configuration target <code>CT</code> is a template engine specific component
+ * that needs to be setup for the engine to function correctly.
+ *
+ * To use the convenience methods in this class, implementations must override {@link #init(Application)}
+ * and call this class' implementation before performing any of their initialization.
+ *
+ * @param <CT> configuration target for custom initialization, implementations could use this
+ *
+ * @see TemplateEngine
+ *
+ * @author Ranganath Kini
+ */
+public abstract class AbstractTemplateEngine<CT> implements TemplateEngine {
+
+    private Languages languages;
+    private Messages messages;
+    private Router router;
+    private PippoSettings pippoSettings;
+
+    private String fileExtension;
+    private String templatePathPrefix;
+
+    /**
+     * Performs common initialization for template engines
+     *
+     * Implementations must override this method to do their own template engine specific initialization. To
+     * use the convenience of this class, implementations must invoke this class's implementation before
+     * performing their own initialization.
+     *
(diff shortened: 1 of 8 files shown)


decebals:
Thanks for contribution. Your idea is good and I think that we can generalize this property, to be used by any template engine. 
So my idea is to add an `AbstractTemplateEngine` (implements `TemplateEngine`) in pippo-core module and to push `fileExtension` variable (and other related code like `setFileExtension` method and the initialization part from `init` method)  in this new class. All template engines (pebble, ...) that exists now in Pippo should be extend this new class.
What you think, make sense?

codematix:
@decebals That sounds useful. It would provide a scaffold for other integrations to build on top of. Do you want me to give it a shot?

decebals:
> Do you want me to give it a shot?

Yes. It's your idea.

codematix:
@decebals Please have a look at the new changes. I have implemented an `AbstractTemplateEngine` and have migrated `PebbleTemplateEngine` to use it. If looks ok, I can migrate the remaining engines to use the new class.

pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * The default implementation throws a {@link NotImplementedException}.
+     *
+     * @param application reference to the Pippo {@link Application} the can be used to access configuration details
+     * @param configurationTarget reference to the template engine specific configuration object that needs initialization
+     *
+     * @throws NotImplementedException
+     *          if no additional initialization is supported by the implementation
+     */
+    protected void init(Application application, CT configurationTarget) {
```
decebals: I don't need this method. In my concrete implementation (`PebbleTemplateEngine` for example) I prefer below approach:
```java
public void init(Application application) {
    super.init(application);

    // specific code ...

    // allow custom initialization
    init(application, builder);
}
```

pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
```diff
@@ -0,0 +1,173 @@
...
+ * To use the convenience methods in this class, implementations must override {@link #init(Application)}
+ * and call this class' implementation before performing any of their initialization.
+ *
+ * @param <CT> configuration target for custom initialization, implementations could use this
+ *
+ * @see TemplateEngine
+ *
+ * @author Ranganath Kini
+ */
+public abstract class AbstractTemplateEngine<CT> implements TemplateEngine {
```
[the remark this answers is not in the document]
codematix: I too did feel this is a bit of an overkill. Will get rid of it.

pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
codematix: Sure makes sense. So if a specific implementation wants to support further extension to the initialization logic, let them provide their own extension point. This will also help get rid of the generic type.

decebals: Yes. You read my mind.

pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    protected void init(Application application, CT configurationTarget) {
+        throw new NotImplementedException();
+    }
+
+    /**
+     * @see TemplateEngine#renderString(String, Map, Writer)
+     */
+    @Override
+    public abstract void renderString(String templateContent, Map<String, Object> model, Writer writer);
```
decebals: No need to override this method here. The method is present already in `TemplateEngine`.

pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * @see TemplateEngine#renderResource(String, Map, Writer)
+     */
+    @Override
+    public abstract void renderResource(String templateName, Map<String, Object> model, Writer writer);
```
decebals: No need to override this method here. The method is present already in TemplateEngine.

codematix: fixed

pippo-core/src/main/java/ro/pippo/core/AbstractTemplateEngine.java
codematix: fixed

codematix:
@decebals I have migrated all `TemplateEngine` implementations to extend from `AbstractTemplateEngine`. Please review and close the PR if it looks good. Thanks!

decebals:
Thanks!

Issue #395, opened 2017-09-27
How to use path parameters encoded in WebSocket url ?


kant111:
currently there seems to be no way to specify path parameters in the WebSocket URL in undertow or jetty. I currently use undertow. 

My use case is I have list of names and I would like to create a generic route that will take "name" as a path parameter so I can just have one handler for the whole list of names I got. 

For example I want to be able to do something like below:

```
addWebSocket("/hello/{name}", (webSocketContext, message) -> {
    try {
        somehow get the value that is passed in for ":name" here
        webSocketContext.sendMessage(message);
    } catch (IOException e) {
        e.printStackTrace();
    }
});
```
To elaborate even further, I am trying to broadcast a messages for all the clients that are listening on specific endpoint. so if I have something like below I wan to be able to broadcast all messages that have a connection opened on "/hello/foo" endpoint. so here my "{name}" = "foo" but I can pass anything from the list of names I have in which case I want to be able to pass name as path parameter.

```
addWebSocket("/hello/foo", (webSocketContext, message) -> {
    try {
        so ":name" = "foo" here
        webSocketContext.sendMessage(message);
    } catch (IOException e) {
        e.printStackTrace();
    }
});
```

decebals:
I will come with a solution.

kant111:
@decebals Appreciate it!!

decebals:
I propose as solution https://github.com/decebals/pippo/pull/397 but I don't know when I have some time to finish the PR.

kant111:
@decebals The commit looks good! What else needs to be done to finish the PR?

decebals:
@kant111 We should: 
- expose the path parameters (we have these information already), add new useful methods in WebSocketContext like `getQueryString()`, `getSession()`, `getParameterNames()`, `getParameterValue()`.
- add DefaultUriMatcherTest
- review new code (classes/interfaces names, integration with existed code - DefaultRouter, ...)

decebals:
The final solution:
```java
addWebSocket("/ws/chat/{roomId}", (webSocketContext, message) -> {
    String roomId = webSocketContext.getPathParameter("roomId").toString();
    System.out.println("roomId = " + roomId);
});
```

I will upgrade [demo](https://github.com/decebals/pippo-demo/blob/master/pippo-demo-websocket/src/main/java/ro/pippo/demo/websocket/WebSocketApplication.java) application very soon.

PR #397, opened 2017-09-28
Add path params in websocket uri

decebals:
This PR contains a functional POC (prof of concept) for #395.
The code works but need some adjustment.

For Undertow:
```java
addWebSocket("/ws/chat/{room}", (webSocketContext, message) -> {
    WebSocketConnection connection = webSocketContext.getConnection();
    // cast WebSocketConnection to UndertowWebSocketConnection
    UndertowWebSocketConnection undertowConnection = (UndertowWebSocketConnection) connection;
    String remoteAddress = undertowConnection.getRemoteAddress().toString();
    System.out.println("remoteAddress = " + remoteAddress);
    WebSocketHttpExchange exchange = undertowConnection.getExchange();
    String query = exchange.getQueryString();
    System.out.println("query = " + query);
});
```

For Jetty the code is similar with the code presented in https://github.com/decebals/pippo/pull/360#issuecomment-296453997.

Also this PR comes with a hidden nice feature, `UriMatcher` with `DefaultUriMatcher`. So, the `DefaultRouter` was spitted in two parts.  It's more easy to implement a custom router with uri matcher (maybe some one wants another notation for path parameters or supply a more fast uri matching algorithm).

In `WebSocketContext` you will can access the parameters value (I have already these information), similar with the approach used in Request. Also, `WebSocketContext` will contain other information like session, request header.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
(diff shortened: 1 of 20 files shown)


kant111:
How to access the value for "room" in the above url? Can we have annotations similar to the HTTP route handling in pippo.

decebals:
> How to access the value for "room" in the above url?

The path parameters are available in [WebSocketMatch](https://github.com/decebals/pippo/pull/397/files#diff-bcced92142528f7f84c34696aaab5c0dR60). The easy mode to expose these parameters is via WebSocketContext (add a method, eventually use ParameterValue as wrapper - to have conversion methods - similar with Request class)

> Can we have annotations similar to the HTTP route handling in pippo.

Do you talk about pippo-controller?

decebals:
@kant111 

> How to access the value for "room" in the above url?

You can use `WebSocketContext.getPathParameter(String name)`.

My work on this feature is done. If everything is OK, I will merge this PR and I will release a new version.

kant111:
@decebals Thanks much! I went with another framework because of deadlines but appreciate your effort! will try to consider Pippo next time when I get an opportunity!

PR #402, opened 2017-10-07
Resolve #400

### LINKED ISSUE #400: Duplicate routes being created
Subtitle: ControllerRegistry.getControllerMethods returning inherited methods

**[1]**
I took unwanted behavior (at least that's what I believe!) in
https://github.com/decebals/pippo/blob/release-1.5.0/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java#L274

The ``Class.getDeclaredMethods()``, which in theory (and by JavaDoc!) should return only the methods of the target class, without returning the inherited methods, is also returning the inherited methods that **are overwritten and use generics**. To solve this I suggest using the method below:

```java
public class ReflectionUtils {
    public static List<Method> getDeclaredMethods(Class clazz) {
        // @formatter:off
        return Arrays
                .stream(clazz.getDeclaredMethods())
                .filter(method -> !method.isBridge())
                .collect(Collectors.toList());
        // @formatter:on
    }
}
```
I noticed this because I have a Controller that inherits from a generic Controller (see below) ...

```java
public abstract class CrudControllerAbstract<ID extends Serializable> {
    protected void delete(ID id) {
    	// ...
    }
}

public class LoggerEntityController extends CrudControllerAbstract<Long> {
    @GET("/logger/delete/{id}")
    public void delete(@Param("id") Long id) {
        super.delete(id);
    }
}
```

... and Pippo has created two routes for the same method. This problem - of creating two routes for the same method - can be solved with the code I suggest above, but also has a connection with a ``strange/different/bizarre`` behavior that I say below in [2].

**[2]**
I saw a different behavior with the same (same same!) JVM version in ``Fedora 25`` and ``RHEL 7`` in the method ``Method.getAnnotations()`` in
https://github.com/decebals/pippo/blob/release-1.5.0/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java#L275

In the loop in ``ControllerRegistry.java#L274``, as I said in [1], the methods ``LoggerEntityController.delete(java.lang.Long)`` and ``LoggerEntityController.delete(java.io.Serializable)`` are iterated. For the method ``LoggerEntityController.delete(java.lang.Long)`` the ``method.getAnnotations()`` returns the `@GET` annotation and all ok here, but for the method ``LoggerEntityController.delete(java.io.Serializable)`` the ``method.getAnnotations()`` has different results in my ``Fedora 25`` and ``RHEL 7``. In ``Fedora 25`` it does not return any annotations and everything is ok here, but in ``RHEL 7`` it returns the `@GET` annotation and here we have a problem since this method is not annotated with `@GET`. In ``RHEL 7``, since both methods are erroneously found to have the `@GET` annotation, then two routes are added for the same method (ControllerRegistry.java#L277: httpMethodAnnotationClasses.contains(annotationClass) is true).

```
java -version
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b31)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b31, mixed mode)
```

**[3]**
A pull request with the ReflectionUtils class [1] is valid and use it on the ControllerRegistry.java#L275 ?

PS: I find it a bit complicated to explain. If I can clarify more, let me know.

mhagnumdw:
Resolve #400 Duplicate routes being created

### DIFF
--- a/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java
+++ b/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java
@@ -271,7 +271,7 @@ public class ControllerRegistry {
         Map<Method, Class<? extends Annotation>> controllerMethods = new LinkedHashMap<>();
 
         // discover all annotated controllers methods
-        for (Method method : controllerClass.getDeclaredMethods()) {
+        for (Method method : ClassUtils.getDeclaredMethods(controllerClass)) {
             for (Annotation annotation : method.getAnnotations()) {
                 Class<? extends Annotation> annotationClass = annotation.annotationType();
                 if (httpMethodAnnotationClasses.contains(annotationClass)) {
--- a/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java
+++ b/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java
@@ -32,6 +32,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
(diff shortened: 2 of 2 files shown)


decebals:
Thanks!

PR #403, opened 2017-10-11
Remove code duplication from ParameterValue

munendrasn:
@decebals 

> Not related to this commit. Sharing it here instead of raising separate issue
```
https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/ParameterValue.java#L247
If values has single entry, 
toSet() handles it differently compared to toList(). Is there any reason for this behavior??

In toSet() method, value is split only on ,(comma) unlike toList(). I think, the behavior should be similar to toList(). 
```


decebals:
Thanks! Good catch. 
I am very happy when I see new unit tests.

PR #412, opened 2018-01-12
Return immutable Set when toSet() is called

munendrasn:
This is to be consistent with toList() which returns immutable List
Extension of PR #403


pippo-core/src/main/java/ro/pippo/core/ParameterValue.java
```diff
@@ -236,15 +236,20 @@ public String toString(String defaultValue) {
     }
 
     public Set<String> toSet() {
-        return toSet(new HashSet<>());
+        return toSet(Collections.emptySet());
     }
 
+    /**
+     *
```
decebals: Can you add a minimal description of method. It's a little bit strange to be empty.
Thanks!

decebals:
Thanks!

PR #413, opened 2018-01-12
Remove braces from start and end in getStrings

munendrasn:
This would help the user to specify properties like below,
`key=[value1, value2, value3]` (more verbose)
Based on 
https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/ParameterValue.java#L277

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/PippoSettings.java
+++ b/pippo-core/src/main/java/ro/pippo/core/PippoSettings.java
@@ -640,7 +640,7 @@ public class PippoSettings {
      * Returns a list of comma-delimited integers from the specified name.
      *
      * @param name
-     * @return list of strings
+     * @return list of integers
      */
     public List<Integer> getIntegers(String name) {
         return getIntegers(name, DEFAULT_LIST_DELIMITER);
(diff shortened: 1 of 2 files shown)


munendrasn:
@decebals 
https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/PippoSettings.java#L496 - This is something I found while I was making this change. What was the reason behind this?? reason for splitting value on `space` and taking the first value and parsing it).

decebals:
@munendrasn Related to your question about:
```java
return Integer.parseInt(value.trim().split(" ")[0]);
```
The code is a contribution of @gitblit (James has a big contribution to Pippo, he's a coauthor). I don't know the reason of split.
The code seems that ignore everything after first space (" "), probably a comment or something like this.
Maybe @gitblit can tell us more. 
If we don't receive a response I think that it's better to delete the split method.

gitblit:
I think the reason was to handle trailing comments.  But I don't remember.  (hi @decebals 😄 )

decebals:
@munendrasn So, we can delete the split method. If someone wants to put a comment he can put it as a comment line.

@gitblit Hi my friend :smile:. I hope you are well.

decebals:
Good work. Thanks!

PR #417, opened 2018-03-28
Add option converterClass to @ParamField

mhagnumdw:
Hi!

The idea is to expand Pippo's ability to convert HTTP request parameters.

With this change it is possible to explicitly a converter to a specific attribute, not limiting the current conversion capacity of the ParameterValue.class.

**I still consider my change a draft, but functional.** I would like to know what you think of this idea and I would also like you to help me make this solution better structured and integrated with the current Pippo code.

The changes I have made are small and I believe the description of the commits is clear.

Here is an example of use:

### Controller
```java
public class DemandController {
    @POST("/demand/save")
    public void save(@Bean Demand demand) {
        super.save(demand);
    }
}
```
### Entity
_New: `convertClass` attribute of annotation `@ParamField`_
```java
public class Demand {
    @ParamField(value = "developer", converterClass = UserSetConverter.class)
    private Set<User> developers = new HashSet<User>();
}
```
### Converter
```java
import ro.pippo.core.Converter;
```
```java
public class UserSetConverter implements Converter {
    @Override
    public Object getAsObject(String[] values) {
        final ParameterValue parameterValue = new ParameterValue(values);
        final List<Long> ids = parameterValue.toList(Long.class);
        // simply instantiate User only with id
        return ids.stream().map(idUser -> new User(idUser)).collect(Collectors.toSet());
    }
}
```
or
```java
public class UserSetConverter implements Converter {
    @Override
    public Object getAsObject(String[] values) {
        final ParameterValue parameterValue = new ParameterValue(values);
        final List<Long> ids = parameterValue.toList(Long.class);
        final UserService userService = GuiceInjector.get().getInstance(UserService.class);
        // using Google Guice retrieves the User from the database
        return ids.stream().map(idUser -> userService.getById(idUser)).collect(Collectors.toSet());
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 Micro Java Web Framework
 =====================
 [![Join the chat at https://gitter.im/decebals/pippo](https://badges.gitter.im/decebals/pippo.svg)](https://gitter.im/decebals/pippo?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Travis CI Build Status](https://travis-ci.org/decebals/pippo.png)](https://travis-ci.org/decebals/pippo)
-[![Coverage Status](https://coveralls.io/repos/decebals/pippo/badge.svg?branch=master&service=github)](https://coveralls.io/github/decebals/pippo?branch=master)
+[![Travis CI Build Status](https://travis-ci.org/pippo-java/pippo.png)](https://travis-ci.org/pippo-java/pippo)
+[![Coverage Status](https://coveralls.io/repos/pippo-java/pippo/badge.svg?branch=master&service=github)](https://coveralls.io/github/pippo-java/pippo?branch=master)
 [![Maven Central](http://img.shields.io/maven-central/v/ro.pippo/pippo.svg)](http://search.maven.org/#search|ga|1|ro.pippo)
 
 <!--
-[![Issue Stats](http://www.issuestats.com/github/decebals/pippo/badge/issue?style=flat)](http://www.issuestats.com/github/decebals/pippo)
-[![Issue Stats](http://www.issuestats.com/github/decebals/pippo/badge/pr?style=flat)](http://www.issuestats.com/github/decebals/pippo)
+[![Issue Stats](http://www.issuestats.com/github/pippo-java/pippo/badge/issue?style=flat)](http://www.issuestats.com/github/pippo-java/pippo)
+[![Issue Stats](http://www.issuestats.com/github/pippo-java/pippo/badge/pr?style=flat)](http://www.issuestats.com/github/pippo-java/pippo)
 -->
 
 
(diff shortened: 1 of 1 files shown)


decebals:
Your idea is not bad. Please give me some time to reflect on it.

decebals:
After a period of thinking, I feel that this concept can be good but not in controllers. In controllers it seems like something magical (I prefer non magical things). With `@Bean` annotation you can materialize a relative simple object hierarchy and probably is enough. 
I don't know if it's not good to have a `ConverterRegistry` in `Application` and to have the possibility to add global converters using something like `setConverter(Double.class, new MyDoubleConverter())` and when I need to convert from a String to Double (anywhere) to use the registered converter. 
Again, for the moment, I don't know where and how to use this concept,

mhagnumdw:
> After a period of thinking, I feel that this concept can be good but not in controllers. In controllers it seems like something magical (I prefer non magical things). With @Bean annotation you can materialize a relative simple object hierarchy and probably is enough.

Passing a Convert to ParameterValue seems like something quite simple and without magic, and provides a great power of conversion.

A simple example: at several points in my code I need to get an [OTRS](https://otrs.com/) number, something like #7146927. I created a type called OTRSTicketNumber to represent this number, its format and even with the #.

So this is much more interesting:
```java
@POST("/exec")
public void exec(@Bean Demand demand) {
    demanda.getOtrs(); // OTRSTicketNumber is assembled! uhu!
}
```

Than this
```java
@POST("/exec")
public void exec(@Bean Demand demand) {
    // Assembling OTRSTicketNumber
    String otrsString = getRequest().getParameter("otrs").toString();
    OTRSTicketNumber otrs = new OTRSTicketNumber(otrsString);
    demanda.setOtrs(otrs);
    demanda.getOtrs(); // OTRSTicketNumber is assembled
}
```

> I don't know if it's not good to have a ConverterRegistry in Application and to have the possibility to add global converters using something like setConverter(Double.class, new MyDoubleConverter()) and when I need to convert from a String to Double (anywhere) to use the registered converter.

Maybe yes. But for generic types I think the simplest way would still be to put the Convert class in the attribute, something loke: `@ParamField(value = "developer", converterClass = UserSetConverter.class)`

Another good option is to register the converters using ServiceLoader, making the class that inherits from Application gets cleaner.

The ParameterValue class is very simple and very useful for converting the most basic types, so I do not think we would need - at least for now - to create Converters for the types already supported by ParameterValue.

Anyway I'm trying to come up with a simple implementation. I know this may not agree with Pippo's principles and ideas and therefore not be accepted (no problem for me), but I think it's something that Pippo could bear.

PS: I did a force-push and unintentionally deleted the history of this PR commits. Sorry!

decebals:
I don't know if I will use this concept very soon, but if it solves for you a practical problem in an elegant way, I say that we can implement it. 
What I wish is to have a light as possible implementation for this concept. That means only one (functional) interface `Converter` and the logical code that test if the `converterClass` parameter is present in annotation and do the conversion if the response is affirmative. I think that it is enough, and we don't need default implementations as `LongConverter`, `FloatConverter`, ...
By the way, I think that we need to add `Locale` as parameter in the `Converter`'s method.

mhagnumdw:
I agree. I'm aligned with you!

I was thinking if we should also pass the `pattern` parameter to the Converter. What do you think?

decebals:
> ParamField has the pattern attribute. Do you think we should pass this attribute to the Converter?

I don't know what to say. Probably is a good move.

mhagnumdw:
@decebals 

I created another branch in my repository to work on it. Since it is not possible to change the source branch of this PR, can I open another PR? Is it alright for you?

decebals:
Sure. Add a reference to this PR to not lose the comments/context.

PR #418, opened 2018-04-09
Convert to float, double and BigDecimal according Locale

mhagnumdw:
`ParameterValue`  when converting numbers does not consider locale. For example: in Brazil the decimal part (fraction) is separated with a `,` (comma), not with a `.` (dot) as in US. Therefore, is not possible submit, for example: `1234,45` or `1.234,56`

In this Pull Request I solved this problem. The `Locale` default is `Locale.getDefault()` and you have the option to pass a specific Locale.

I have adapted the tests and added more tests.

Please, tell me if it is all fine and if any change is necessary.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/ParameterValue.java
+++ b/pippo-core/src/main/java/ro/pippo/core/ParameterValue.java
@@ -24,6 +24,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.math.BigDecimal;
 import java.sql.Time;
 import java.sql.Timestamp;
+import java.text.DecimalFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


decebals:
Can you show me how you use Locale in your code with this PR? Where you pass Locale in your code?

mhagnumdw:
Hi, Decebals!

Well, for this PR, the goal is turn ParameterValue locale-sensitive. Now, a user-agent that submit a number formated with pt-BR it will be converted correctly if the JVM Locale is pt-BR. Same behavior for others locales. I belive this change is backward compatible.

Next step, maybe in another PR, Pippo can set Locale in ParameterValue according `Accept-Language` provided by `HttpServletRequest.getLocale()`.

What do you think?

decebals:
> Next step, maybe in another PR, Pippo can set Locale in ParameterValue according Accept-Language provided by HttpServletRequest.getLocale().

That's what I was thinking about. With this PR we can set any local in parameter value, but actually we don't use this feature because only in `Response` we create `ParameterValue` objects (now we use the constructor without parameters).

decebals:
Thanks!

PR #419, opened 2018-04-11
Uses Locale in request context

mhagnumdw:
This PR is a part of the PR #418. Now request context pass Locale from `Accept-Language` Header to `ParameterValue`, making flexible the conversion according to location.

Tests included!

Some things:
- Do you think need create a `getLocale()` in `RouteContext` obtaining the value from `Request.getLocale()`?
- The classes `WebSocketContext` and `CsvEngine` make use of `ParameterValue`, but I did not make changes because there is no request context.

Please, let me know about **any changes** you want me to make. Even class names, methods names, class locations and writing!


decebals:
I think that integration tests (via PippoTest and Controller) are oversize. It's enough to add new test methods with Locale in ParameterValueTest.

pippo-core/src/main/java/ro/pippo/core/Request.java
```diff
@@ -391,6 +392,19 @@ public String getContentType() {
         return contentType;
     }
 
+    /**
```
decebals: It's enough as javadoc something like: "It's wrapper over {@link HttpServletRequest#getLocale}."

decebals:
> Do you think need create a getLocale() in RouteContext obtaining the value from Request.getLocale()?

It's not needed a such method in RouteContext. In RouteContext we put which is often used.

> The classes WebSocketContext and CsvEngine make use of ParameterValue, but I did not make changes because there is no request context.

OK. I will take a look.

mhagnumdw:
@decebals 
Do you think I delete ControllerLocaleTest or I let a little test using `Accept-Language` Header ?

decebals:
> Do you think I delete ControllerLocaleTest or I let a little test using Accept-Language Header ?

We need only to add new tests for the methods that deal with `Locale` from `ParameterValue`, so it's enough to add some new test methods in existed `ParameterValueTest` class. The locale object passed to ParamaterValue, in real world, is took via `Request.getLocale` method that calls `HttpServletRequest.getLocale` (it's a simple wrapper, a banal shortcut), so we don't need to test `Accept-Language` Header, it's already done by `HttpServletRequest.getLocale`.

Make sense for you?

In the feature I thing that it's a good idea to debate a new feature in a new issue, and after we come to a conclusion, we can create a PR to present and comment an implementation. The scope is to not spent time in a wrong direction.

mhagnumdw:
> Make sense for you?

Yes!

> In the feature I thing that it's a good idea to debate a new feature in a new issue, and after we come to a conclusion, we can create a PR to present and comment an implementation. The scope is to not spent time in a wrong direction.

This is really better.

pippo-core/src/main/java/ro/pippo/core/Request.java
mhagnumdw: Done!

pippo-controller-parent/pippo-controller/pom.xml
```diff
@@ -34,6 +34,22 @@
             <version>${servlet.version}</version>
             <scope>provided</scope>
         </dependency>
+
```
decebals: I think that this file remains unmodified.

pippo-core/src/test/java/ro/pippo/core/ParameterValueTest.java
```diff
@@ -137,6 +158,27 @@ public void testDouble() throws Exception {
         assertArrayEquals(new Double[]{3.14159d, 3.14159d, 3.14159d}, new ParameterValue(US, "3.14159", "3.14159", "3.14159").to(Double[].class));
     }
 
+    @Test
+    public void testDoublePtBR() throws Exception {
+    	final Locale PT_BR = new Locale("pt", "BR");
```
decebals: Indentation

pippo-core/src/test/java/ro/pippo/core/ParameterValueTest.java
```diff
@@ -160,15 +202,36 @@ public void testDoubleDefaultLocale() throws Exception {
...
+        assertEquals(new BigDecimal(0d), new ParameterValue(US, "").toBigDecimal());
+        assertEquals(new BigDecimal(0d), new ParameterValue(US, " ").toBigDecimal());
         assertEquals(new BigDecimal("3.14159"), new ParameterValue(US, "3.14159").toBigDecimal());
         assertEquals(new BigDecimal("3.14159"), new ParameterValue(US, "3.14159", "3.14159", "3.14159").toBigDecimal());
         assertArrayEquals(new BigDecimal[]{new BigDecimal("3.14159"), new BigDecimal("3.14159"), new BigDecimal("3.14159")}, new ParameterValue(US, "3.14159", "3.14159", "3.14159").to(BigDecimal[].class));
     }
 
+    @Test
+    public void testBigDecimalPtBR() throws Exception {
+    	final Locale PT_BR = new Locale("pt", "BR");
```
decebals: Indentation

pippo-core/src/test/java/ro/pippo/core/ParameterValueTest.java
mhagnumdw: Done!

pippo-core/src/test/java/ro/pippo/core/ParameterValueTest.java
mhagnumdw: Done!

decebals:
Thanks!

PR #431, opened 2018-06-17
Updated the right dependencies

jpsingarayar:
The dependencies for the jetty server is missing hence added the same.


decebals:
Thanks!

PR #433, opened 2018-06-18
FileItem's inputstream is closed properly now

apoorvprecisely:
write method of file item doesn't closed newly created `inputStream` post usage
this fix closes it in a finally block


apoorvprecisely:
we faced an issue where file descriptors were not getting released on file uploads.
Requires process restarts to do that

decebals:
I changed try-catch-finally with a try-with-resources block in your code.
Good catch. Thanks!

PR #434, opened 2018-06-26
Add Handlers class as a factory for common RouteHandler

decebals:
The idea is that sometime I want to have a place with all common route handlers (defined in core). 
It's something similar with Executors from Java.
The implementation based on factory pattern.

What you say, it's useful or not?


decebals:
In the end I don't know if it's so useful in the context of IDE because you can write 
```java
new Handle
```
and if you press CTRL-SPACE, the Intellij Idea shows you all available handler classes.

mhagnumdw:
Apparently not much needed. In Eclipse Ctrl + T shows all the implementations.

PR #436, opened 2018-07-01
The Settings File is now read with UTF-8 as the default encoding.

rygel:
I've encountered a problem where the strings read from PippoSettings were garbling German Umlauts. This is because the File is not opened as a UTF-8 file, but instead with the local encoding. This PR now opens the Settings File always as UTF-8.

As an alternative I've tested supplying -Dfile.encoding=UTF8 to the JVM instead the pathc, but it is not helping.


decebals:
Thanks!
I forgotten about this PR. Another time, please remember me :smile:.

rygel:
Sorry. I thought you were very busy. Next time I will remind you :)

PR #437, opened 2018-07-16
Add useful handlers for admin

decebals:
I added small route handlers that are useful in my opinion for an admin.

```java
GET("/ping", new PingHandler());
GET("/threadDump", new ThreadDumpHandler());
GET("/metrics", new MetricsHandler());
GET("/healthCheck", new HealthCheckHandler(healthCheckRegistry));
GET("/settings", new SettingsHandler());
GET("/sysInfo", new SystemInfoHandler());
```

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/route/SystemInfoHandler.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/SystemInfoHandler.java
@@ -45,11 +45,16 @@ public class SystemInfoHandler implements RouteHandler {
     public void handle(RouteContext routeContext) {
         Response response = routeContext.getResponse().noCache().text();
 
+        Map<String, String> props = new HashMap<>();
+        for (String name : System.getProperties().stringPropertyNames()) {
+            props.put(name, System.getProperty(name));
+        }
+
         try (BufferedWriter writer = new BufferedWriter(response.getWriter())) {
-            writeSystemProperties(writer);
+            writeSystemProperties(props, writer);
             writer.newLine();
 
-            writeEnv(writer);
+            writeEnv(System.getenv(), writer);
 
             writer.flush();
         } catch (IOException e) {
@@ -57,19 +62,14 @@ public class SystemInfoHandler implements RouteHandler {
         }
     }
 
-    protected void writeSystemProperties(BufferedWriter writer) throws IOException {
+    protected void writeSystemProperties(Map<String, String> props, BufferedWriter writer) throws IOException {
         writeBanner("System properties", writer);
-
-        Map<String, String> props = new HashMap<>();
-        for (String name : System.getProperties().stringPropertyNames()) {
-            props.put(name, System.getProperty(name));
-        }
         writeProperties(props, writer);
     }
 
-    protected void writeEnv(BufferedWriter writer) throws IOException {
+    protected void writeEnv(Map<String, String> env, BufferedWriter writer) throws IOException {
         writeBanner("Environment", writer);
-        writeProperties(System.getenv(), writer);
+        writeProperties(env, writer);
     }
 
     protected void writeBanner(String banner, BufferedWriter writer) throws IOException {
--- a/pippo-core/src/main/java/ro/pippo/core/route/SystemInfoHandler.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/SystemInfoHandler.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ro.pippo.core.route;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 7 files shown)


pippo-metrics-parent/pippo-metrics/src/main/java/ro/pippo/metrics/ThreadDumpHandler.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void handle(RouteContext routeContext) {
+        Response response = routeContext.getResponse().noCache().text();
+
+        if (threadDump != null) {
+            OutputStream output = response.getOutputStream();
+            threadDump.dump(output);
+        } else {
+            response.internalError().send("Sorry your runtime environment does not allow to dump threads");
```
mhagnumdw: `internalError()` return 500, not 501. JavaDoc says

> If the thread dump is not supported then the status code returned is 501 (not implemented).

pippo-core/src/main/java/ro/pippo/core/route/SystemInfoHandler.java
```diff
@@ -0,0 +1,74 @@
...
+        Map<String, String> props = new HashMap<>();
+        for (String name : System.getProperties().stringPropertyNames()) {
+            props.put(name, System.getProperty(name));
+        }
+
+        Map<String, String> env = System.getenv();
+
+        Response response = routeContext.getResponse().noCache().text();
+
+        try (BufferedWriter writer = new BufferedWriter(response.getWriter())) {
```
mhagnumdw: What do you think of differentiating in output what is system property and what is environment variable?

mhagnumdw:
> GET("/metrics", new MetricsHandler());

I did not find the class MetricsHandler.

Overall the idea seems to be cool!

pippo-metrics-parent/pippo-metrics/src/main/java/ro/pippo/metrics/ThreadDumpHandler.java
decebals: I made the correction.

decebals:
In the future I want to add query parameter to these handlers (where is applicable) to filter some information. For example I want to retrieve only the system properties from `SystemInfoHandler` so, I will use "sysprops" as parameter.

Another idea is to adapt somehow the content type of the responses. Now these handlers return text only (maybe I want an `html` table or a `json`). From this reason I introduce the protected method, to have a possibility to "format" the content/result. It's only an idea and I am not sure that is valuable.

pippo-core/src/main/java/ro/pippo/core/route/SystemInfoHandler.java
mhagnumdw: This approach is good.

decebals:
> I did not find the class MetricsHandler.

This handler is not prod ready. The idea is that it's relative complex to convert a [MetricRegistry](https://metrics.dropwizard.io/3.1.0/apidocs/com/codahale/metrics/MetricRegistry.html) to plain text.
We can use [MetricsModule](https://metrics.dropwizard.io/3.1.0/apidocs/com/codahale/metrics/json/MetricsModule.html) to convert in `json` but this solution comes with a dependency to [metrics-json](https://mvnrepository.com/artifact/com.codahale.metrics/metrics-json). To convert metrics to json using metrics-json library can be, in the end, the correct solution (or at least, an acceptable solution).

decebals:
I added `MetricsHandler`.

See below some screenshots of this admin handlers:

**SettingsHandler**
image

**SystemInfoHandler**
image

**ThreadDumpHandler**
image

**MetricsHandler**
image

decebals:
See below how you can add a new `HealthCheck` implementation (`DatabaseHealthCheck`):

```java
import com.codahale.metrics.health.HealthCheck;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import javax.sql.DataSource;

import java.sql.Connection;
import java.sql.DatabaseMetaData;
import java.sql.SQLException;
import java.sql.Statement;

public class DatabaseHealthCheck extends HealthCheck {

    private static final Logger log = LoggerFactory.getLogger(DatabaseHealthCheck.class);

    private final DataSource dataSource;
    private final String healthCheckSql;

    private final String healthyMessage;
    private final String unHealthyMessage;

    public DatabaseHealthCheck(DataSource dataSource, String healthCheckSql) throws SQLException {
        this.dataSource = dataSource;
        this.healthCheckSql = healthCheckSql;

        DatabaseMetaData metaData = dataSource.getConnection().getMetaData();
        String url = metaData.getURL();
        String userName = metaData.getUserName();

        healthyMessage = String.format("Database connection successful (host = %s, username %s", url, userName);
        unHealthyMessage = String.format("Unable to connect to database (host = %s, username %s'", url, userName);
    }

    @Override
    protected Result check() {
        Result result;

        try (Connection connection = dataSource.getConnection(); Statement statement = connection.createStatement()) {
            boolean ok = statement.execute(healthCheckSql);
            result = ok ? Result.healthy(healthyMessage) : Result.unhealthy(unHealthyMessage);
        } catch (Exception e) {
            log.error(e.getMessage(), e);
            result = Result.unhealthy(unHealthyMessage);
        }

        log.debug(result.getMessage());

        return result;
    }

}
```

and now you must register the new health check implementation:
```java
import com.codahale.metrics.health.HealthCheckRegistry;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class MyApplication extends Application {

    private static final Logger log = LoggerFactory.getLogger(MyApplication.class);

    private final HealthCheckRegistry healthCheckRegistry;

    public MyApplication() {
        healthCheckRegistry = new HealthCheckRegistry();
        try {
            healthCheckRegistry.register("database", new DatabaseHealthCheck(DbUtils.getDataSource(), "SELECT 1 FROM dual"));
        } catch (SQLException e) {
            log.error(e.getMessage(), e);
        }
    }

    @Override
    protected void onInit() {
        GET("/healthCheck", new HealthCheckHandler(healthCheckRegistry));
    }

}
```

PR #446, opened 2018-09-17
feat: upgrade guice to latest version

munendrasn:
* With this version, the multi-binding extension would be part of
  main Guice package
@decebals Please review


decebals:
Thanks!

What means 

> With this version, the multi-binding extension would be part of main Guice package

is it valuable? (I don't use Guice in my projects, so sorry me for ignorance)

Issue #455, opened 2018-09-28
Check if stream is closed


kiru:
Hello, 

While debugging, I ended up several times [here](https://github.com/pippo-java/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseStream.java#L47) which was called by [this](https://github.com/pippo-java/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/util/IoUtils.java#L134). 

We can avoid this by check if the stream is already closed or not. 

Thanks, 
Kiru

decebals:
Sure. Please submit a PR.

decebals:
Closed by #457

PR #457, opened 2018-09-29
Closing the writer will close the stream as well. Fix for #455.

kiru:
Calling close twice on the stream will throw an exception ( which is silently ignored ). 
I don't think this change has any effect. It is just annoying if you debug your application and set the debugger to halt on every exception and see it halting within Pippo.

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseWrapper.java
+++ b/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseWrapper.java
@@ -81,7 +81,6 @@ public class GZipResponseWrapper extends HttpServletResponseWrapper {
 
     public void finish() {
         IoUtils.close(writer);
-        IoUtils.close(stream);
     }
 
     private ServletOutputStream createOutputStream() throws IOException {


decebals:
> Calling close twice on the stream will throw an exception ...

Can you explain why twice?

kiru:
> > Calling close twice on the stream will throw an exception ...
> 
> Can you explain why twice?

Calling getWriter() on a new instance of GZipResponseWrapper will [create a new stream ](https://github.com/pippo-java/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseWrapper.java#L71)and [new *PrintWriter* wrapping that stream](https://github.com/pippo-java/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseWrapper.java#L72).

Now GZipResponseStream is implemented such that it [throws an exception if you call close() twice](https://github.com/pippo-java/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseStream.java#L47). 

The finish methods in GZipResponseWrapper calls close on the **writer** first and then on the **stream**.  Now calling close on the **writer** will close the underlying **stream**.  Another call on **stream** will cause an IOException.

IMHO, I don't think ignoring the exception returned by [IOUtils.close()](https://github.com/pippo-java/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseWrapper.java#L83) is a good idea, I'd rather add a not-null check there.

decebals:
@kiru Thanks! Sorry for the delay.

> IMHO, I don't think ignoring the exception returned by IOUtils.close() is a good idea, I'd rather add a not-null check there.

Now I think also that `IOUtils.close` is more or less useless and it can lead to problems. We can remove it. A PR is welcome.

PR #463, opened 2018-10-06
fix(undertow): resolve undertow settings not getting picked

### LINKED ISSUE #462: Undertow Settings not being picked from settings
As part of `1.10.0` pippo release,  undertow version was upgraded to `1.4.25`.  With this version, undertow added strict [HTTP parsing](https://github.com/undertow-io/undertow/commit/09d40a13089dbff37f8c76d20a41bf0d0e600d9d). This can overriden if needed by setting `ALLOW_UNESCAPED_CHARACTERS_IN_URL` to `true`

Pippo supports pass undertow options value in settings. So, I added `undertow.server.ALLOW_UNESCAPED_CHARACTERS_IN_URL=true` but this was not picked up undertow server.
On debugging, it was because of the way pippo added `Option` from propertyName, [here](https://github.com/pippo-java/pippo/blob/master/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowSettings.java#L186)

munendrasn:
Fixes #462 
This has been tested with an independent project.

### DIFF
--- a/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowSettings.java
+++ b/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowSettings.java
@@ -22,6 +22,7 @@ import org.slf4j.LoggerFactory;
 import org.xnio.Option;
 import ro.pippo.core.PippoSettings;
 import ro.pippo.core.WebServerSettings;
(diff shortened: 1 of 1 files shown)


decebals:
Thanks!

I think that it's a good idea to add some lines about this feature in [documentation](https://github.com/pippo-java/pippo.github.io). If you few minutes, please create a PR in documentation project

Issue #454, opened 2018-09-28
Lead to RCE when unmarshal xml data with XStream


idealzh:
The XstreamEngine unmarshal xml data based on XStream,but it doesn't check the data;
when the data contains malicious types,then it may leads to remote code execution;
The Struts2 framework once had the same issue(https://cwiki.apache.org/confluence/display/WW/S2-052); 
Using the following code snippet to convert the malicious xml data；
     `   POST("/xml",routeContext ->{
        	String xml_data=routeContext.getRequest().getBody();
        	XstreamEngine engine=new XstreamEngine();
        	engine.fromString(xml_data, String.class);
        });`
 The malicious xml data is as follows： 
[payload.txt](https://github.com/pippo-java/pippo/files/2427802/payload.txt)
The tool marshalsec(https://github.com/mbechler/marshalsec) can help to generate more kinds of payload including the one above;
To mitigate the vulnerability,since version 1.4.7,XStream provides developers with some APIs(such as:addPermission,denyPermission,allowTypes,denyTypes) to restrict the types be unmarshalled；
Here we could fix the issue refer to the [patch](https://github.com/apache/struts/commit/19494718865f2fb7da5ea363de3822f87fbda264) of Struts2.

decebals:
@idealzh 
Thanks! I understand the problem. Do you have a solution, or a recommendation for other users that use pippo-xtream? Do you think that we can add a protection in Pippo? If yes, please submit a PR.

decebals:
@idealzh 
If you have questions about Pippo, we are here to respond you.

mhagnumdw:
This issue may have a point in common with issue #458.

rygel:
@mhagnumdw Thank you for explaining the issue better in the comments of #470 . I understand now that both #458 and #454 are related. I am now just wondering this issue is fixed with #470 or if there is more work needed?

mhagnumdw:
@rygel , thank you, I hope to help when possible!

From what I understand this issue has not been resolved. But, looking quickly, the method [XstreamEngine.xstream()](https://github.com/pippo-java/pippo/blob/c6b26551a82d2dd32097fcb17c13c3b830916296/pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java#L42) can make use of the method `allowTypes` from XStream. Other security-related methods are listed here: http://x-stream.github.io/security.html#framework

[...]

PR #470, opened 2018-10-25
Fix 458

rygel:
I've added the fix propsed by @idealzh


rygel:
Hm, I need to check the failure in the tests.

decebals:
I tried to resolve the conflict from merge with master from GitHub but I see that without success.
So please make a merge manually.

What I don't like with this kind of solution is that we modified the core for a vulnerability that appear in a optional module (for example I don't use that module in my projects). And the modification is not trivial. For a new comer is hard to read this code and to quess why we did it. I don't understand why XStream doesn't come with a fix.

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/FilteringObjectInputStream.java
```diff
@@ -0,0 +1,25 @@
+package ro.pippo.session;
```
decebals: Add copyright (copy/paste from other file)

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/ClassFilter.java
```diff
@@ -0,0 +1,24 @@
+package ro.pippo.session;
+
+import java.util.ArrayList;
+
+/**
+ * @author idealzh
+ */
+public class ClassFilter {
```
decebals: We don't need this class. Put this logic directly in `FilteringObjectInputStream`.

rygel:
I had some difficulties while locally merging. ~~I'll check when Travis is finished.~~ Should work now.

About your reservations:
- I am not sure if this issue is even related to XStream. I think that is another isssue? Or I do not understand the implications properly.
- I was implementing this fix, because I don't want to have an open CVE for Pippo. I get all these warnings from Github in my projects which use Pippo.
- I understand that fixing the bug in core is complicating things. If you want you can reject the fix.
- I personally don't use any of these modules.

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/ClassFilter.java
```diff
@@ -0,0 +1,24 @@
+package ro.pippo.session;
+
+import java.util.ArrayList;
+
+/**
+ * @author idealzh
+ */
+public class ClassFilter {
+    private ArrayList<String> WhiteList= null;
```
decebals: Rename variable in "whiteClassNames".

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/FilteringObjectInputStream.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * @author idealzh
+ */
+public class FilteringObjectInputStream extends ObjectInputStream {
+    public FilteringObjectInputStream(InputStream in) throws IOException {
+        super(in);
+    }
+
+    protected Class<?> resolveClass(java.io.ObjectStreamClass descriptor) throws ClassNotFoundException, IOException {
```
decebals: `ObjectStreamClass` is enough, no need for package (it's already imported).

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/FilteringObjectInputStream.java
```diff
@@ -0,0 +1,25 @@
...
+        String className = descriptor.getName();
+        ClassFilter classFilter = new ClassFilter();
+        if(className != null && className.length() > 0 && !classFilter.isWhiteListed(className)) {
```
decebals: I think that we have a method in StringUtils that check if a string is not empty (not null included).

decebals:
@rygel
> I am not sure if this issue is even related to XStream.

Yes, you are right.

>If you want you can reject the fix.

No, I will merge in the end. Maybe I will read more about this subject. Users using Pippo need to make sure we focus on security.

rygel:
I have changed all your requests.

mhagnumdw:
I read and refreshed my memory! :)

Serialization / Deserialization is very common and checking whether certain types may or may not be deserialized is very important because of security concerns. Regardless if it is from a json, xml, Java serialization etc.

Recently (others not so) some frameworks that I use have implemented a restriction to allow or not to deserialize a set of classes and/or packages, eg: [Richfaces 3.3.3 > 3.3.4](http://anonsvn.jboss.org/repos/richfaces/branches/community/release-3.3.4/framework/impl/src/main/java/org/ajax4jsf/resource/LookAheadObjectInputStream.java), [JBoss EAP 6.3.0 with jackson-mapper-asl 1.9.9.redhat-3 > JBoss EAP 6.4.20 with jackson-mapper-asl 1.9.9.redhat-6](https://stackoverflow.com/questions/50917932/what-versions-of-jackson-are-allowed-in-jboss-6-4-20-patch#51329422), XStream after version 1.4.6. So I had to define a set of my allowed packages for Serialization / Deserialization.

I think somewhere in Pippo there could be the option where it would be possible to define a list of classes and/or packages that would be allowed to deserialize (whitelist). I believe somewhere in the core. Why the core? Because, as I said above, Serialization / Deserialization is something common.

This whitelist would already come with some basic classes (eg ro.pippo.core.Flash, java.util.ArrayList etc), but would be extensible in some way (eg: system property, application.properties etc) so that it could be possible to add more classes and/or packages.

Both the FilteringObjectInputStream class (created here in this PR) and the [XstreamEngine](https://github.com/pippo-java/pippo/blob/master/pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java) class (see issue #454), as well as any other Pippo class that works with Serialization / Deserialization, **should check the whitelist!**

Below are two implementations from where we can extract one or another idea (let's always keep in mind the simplicity, efficiency and code readability of Pippo!):
- [BeanDeserializerFactory class in jackson-mapper-asl-1.9.9.redhat-6.jar](https://maven.repository.redhat.com/techpreview/all/org/codehaus/jackson/jackson-mapper-asl/1.9.9.redhat-6/jackson-mapper-asl-1.9.9.redhat-6-sources.jar)
- [LookAheadObjectInputStream class in richfaces4/core](https://github.com/richfaces4/core/blob/master/impl/src/main/java/org/richfaces/util/LookAheadObjectInputStream.java)

What do you guys think?

decebals:
I find this interesting [article](https://tersesystems.com/blog/2015/11/08/closing-the-open-door-of-java-object-serialization) that describes problem in detail.

decebals:
I see some improvements in code. I will add new comments in a new review.

CHANGELOG.md
```diff
@@ -5,9 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### [Unreleased][unreleased]
 
 #### Fixed
+- [#458]: Java deserialization vulnerability in SerializationSessionDataTranscoder.decode()
```
decebals: To have a nice link is necesary to declare the link of `[#ABC]:` to the end of file (it's a list with all issues references).

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/FilteringObjectInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.ObjectInputStream;
+import java.io.ObjectStreamClass;
+import java.util.ArrayList;
+
+import static ro.pippo.core.util.StringUtils.isNullOrEmpty;
+
+/**
+ * @author idealzh
+ */
+public class FilteringObjectInputStream extends ObjectInputStream {
```
decebals: Rename class to `WhitelistObjectInputStream` and move it in `pippi.core.util` package.

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/FilteringObjectInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+    private ArrayList whiteClassNames = new ArrayList<String>();
```
decebals: The type of variable must be `List`.

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/FilteringObjectInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+        String className = descriptor.getName();
+        if (isNullOrEmpty(className) && !isWhiteListed(className)) {
+            throw new InvalidClassException("Unauthorized deserialization attempt", descriptor.getName());
+        } else {
+            return super.resolveClass(descriptor);
+        }
+    }
+
+    private boolean isWhiteListed(String className) {
+        if (className == null) return false;
```
decebals: We don't need this check, the check is already done in `resolveClass` method.

pippo-session-parent/pippo-session/src/main/java/ro/pippo/session/FilteringObjectInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @author idealzh
+ */
+public class FilteringObjectInputStream extends ObjectInputStream {
+    private ArrayList whiteClassNames = new ArrayList<String>();
+
+    public FilteringObjectInputStream(InputStream in) throws IOException {
+        super(in);
+
+        whiteClassNames.add("ro.pippo.session.DefaultSessionData");
```
decebals: We can read these information from a file "src/main/resources/pippo/withelist-serialization.txt" where the content of that file is like (one class name on one line - it's easy to read):
```
ro.pippo.session.DefaultSessionData
java.util.HashMap
ro.pippo.core.Flash
java.util.ArrayList
```

For a reference implementation see [this](https://github.com/richfaces4/core/blob/master/impl/src/main/java/org/richfaces/util/LookAheadObjectInputStream.java).

rygel:
@mhagnumdw Thank you for explaining the issue better. I understand now that both #458 and #454 are related. I am now just wondering if we fixed also the XStream issue with this PR or if there is more work needed?

decebals:
We can improve the format of `whitelist-serialization` file.
For example we can support for a whole package with a line like (or other simple and better format):
```
java.lang.
```
could mean add all classes from `java.lang` package. To add support for package is simple, just create a new `whitePackageNames` (similar with whiteClassNames) and in `isWhiteListed` method add somethig like (pseudocode):
```java
for (String package: whitePackageNames) {
    if (className.startWith(package)) { // trivial implementation
       return true;
    }
}
```

In the extreme case (last step), we can go (add support) with regex. So, the sky is the limit :smile:.

It's not very clear for me what about primitive classes, if we must add them by default in `whitelist-serialization`.

PR #471, opened 2018-10-29
Add whole package to whitelist

rygel:
Add the suggestion to whitelist whole package names in discussion to #470


rygel:
I've added regular expressions now. Which is more powerful than this.

PR #472, opened 2018-10-29
Fix 454

rygel:
This is my take on fixing this bug. I've added regular expression parsing and re-using WhitelistObjectInputStream.

### DIFF
--- a/pippo-content-type-parent/pippo-xstream/pom.xml
+++ b/pippo-content-type-parent/pippo-xstream/pom.xml
@@ -15,7 +15,7 @@
     <description>XStream integration</description>
 
     <properties>
-        <xstream.version>1.4.9</xstream.version>
+        <xstream.version>1.4.11.1</xstream.version>
     </properties>
 
     <dependencies>
--- a/pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java
+++ b/pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java
@@ -21,6 +21,9 @@ import ro.pippo.core.ContentTypeEngine;
 import ro.pippo.core.HttpConstants;
 
 import com.thoughtworks.xstream.XStream;
+import ro.pippo.core.util.WhitelistObjectInputStream;
+
+import java.util.regex.Pattern;
 
 /**
  * An XmlEngine based on XStream.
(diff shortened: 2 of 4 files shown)


rygel:
@decebals Please don't merge the commit cc69274. This is unfinished and should not be part of this branch.

mhagnumdw:
@rygel 
> @decebals Please don't merge the commit [cc69274](https://github.com/pippo-java/pippo/commit/cc6927492f9a42123f7bbf15704868d51d6e24cd). This is unfinished and should not be part of this branch.

It may be better to remove the commit from your branch, so we avoid undue merge.

I will do again the review I did in closed PR #471

rygel:
@mhagnumdw The last commit should have fixed this now.
Thank you for doing your review!

decebals:
I made the modifications.
Maybe now this aspect with security and deserialize is a little bit paranoiac :smile:.
I don't have a Pippo application with XStream to see if something works.

decebals:
I modified the [pippo-demo-basic](https://github.com/pippo-java/pippo-demo/tree/master/pippo-demo-basic), to use xstream instead of jackson (add `pippo-xstream` dependency in front of of `pippo-jackson`)
```xml
<dependency>
    <groupId>ro.pippo</groupId>
    <artifactId>pippo-xstream</artifactId>
    <version>${pippo.version}</version>
</dependency>

<dependency>
    <groupId>ro.pippo</groupId>
    <artifactId>pippo-jackson</artifactId>
    <version>${pippo.version}</version>
</dependency>
```

and I see that `http://localhost:8338/xml` works but my expectation was to not work, because `Contact` class is not in whitelist.
On the other part, to add each domain class (`Contact` in demo) in `WhitelistObjectInputStream` seems (maybe) to a little bit too much (in my opinion). Or maybe it's the normal workflow if you want a secure application. In the end you can use the regex to add all your domain classes (`ro.pippo.demo.*` in demo case).

rygel:
> I modified the [pippo-demo-basic](https://github.com/pippo-java/pippo-demo/tree/master/pippo-demo-basic), to use xstream instead of jackson (add `pippo-xstream` dependency in front of of `pippo-jackson`)
> and I see that `http://localhost:8338/xml` works but my expectation was to not work, because `Contact` class is not in whitelist.

@decebals So no exception has been thrown? I just checked the XStream documentation (http://x-stream.github.io/security.html) and I think I forgot a crucial part, while working on the fix.
```java
// clear out existing permissions and set own ones
xstream.addPermission(NoTypePermission.NONE);
```
I think XStream already is setup with some default permissions and to get the highest securit thoe have to be cleared first, before adding the own/new ones. Sorry for my oversight. I created a new PR for this.

PR #474, opened 2018-11-01
feat(gzip): send gzip response when Accept-Encoding:* is present

### LINKED ISSUE #473: [Question] Gzip With Accept-Encondig:*
When the request has explicit `Accept-Encoding:*`, it would be nice if the gzipped response is sent. With the explicit `*`, it shows that the client is able to handle the gzip response

Also, should Pippo support `qvalue`?? In future, this can be used to support other Encoding. Also, `gzip;q=0` is sent then, gzip response shouldn't be sent .
This is not a problem/issue. As per RFC-7231 
```
Note: Most HTTP/1.0 applications do not recognize or obey qvalues
associated with content-codings.  This means that qvalues might
not work and are not permitted with x-gzip or x-compress.
```

munendrasn:
* When explicit * is specified in Accept-Encoding then send
  gzip response
Fixes #473

@decebals This PR doesn't include `qvalue` changes


decebals:
@munendrasn 
Perfect. Thanks!
I like to see the new unit tests :smile:.

PR #475, opened 2018-11-01
Add support for Deflate Encoding

munendrasn:
As part of this,
* Gzip package has been restructured
In future,
* Support other Encodings
* Support for qvalue in Encoding


decebals:
The PR is good. I see some improvements that I can address with a review.

What is not clear for me is if we need to do it our self. My expectation is that, somehow, the servlet container (Jetty, Undertow, Tomcat) to deals already with this aspect (use a compressed `HttpServletResponse` when a request requires this).
So, we can check it with a small test.

decebals:
I see in Jetty 9 [documentation](https://www.eclipse.org/jetty/documentation/9.4.12.v20180830/gzip-filter.html) that:

> Jetty 9 only compresses using GZip. Using deflate HTTP compression is not supported and will not function.

So, in the end my question is, we need this flexibility (configurable compress method) out of the box? Is it not enough the gzip compression? I don't want to add extra code (and maintain it) without a clear benefit.

rygel:
I am not sure in what configurations Pippo is used by our users. However, for my installations I am using NGINX as a reverse proxy, which handles all SSL and GZIP details.

munendrasn:
Baesd on above discussion closing this

PR #477, opened 2018-11-01
Clear XStream default whitelist

rygel:
See my comment on #472.

### DIFF
--- a/pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java
+++ b/pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java
@@ -15,6 +15,7 @@
  */
 package ro.pippo.xstream;
 
+import com.thoughtworks.xstream.security.NoTypePermission;
 import org.kohsuke.MetaInfServices;
 import ro.pippo.core.Application;
 import ro.pippo.core.ContentTypeEngine;
@@ -48,6 +49,9 @@ public class XstreamEngine implements ContentTypeEngine {
         xstream.autodetectAnnotations(true);
         // prevent xstream from creating complex XML graphs
         xstream.setMode(XStream.NO_REFERENCES);
+
+        // clear out existing permissions and set own ones
+        xstream.addPermission(NoTypePermission.NONE);
 
         //setup security
         xstream.allowTypes((String[]) WhitelistObjectInputStream.getWhitelistedClassNames().toArray());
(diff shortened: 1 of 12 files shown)


pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java
```diff
@@ -41,17 +45,17 @@ public String getContentType() {
...
         xstream.setMode(XStream.NO_REFERENCES);
 
-        // setup security (see http://x-stream.github.io/security.html)
-        xstream.allowTypes(WhitelistObjectInputStream.getWhiteClassNames());
-        xstream.allowTypesByRegExp(WhitelistObjectInputStream.getWhiteRegEx());
+        // clear out existing permissions and set own ones
+        xstream.addPermission(NoTypePermission.NONE);
 
+        //setup security
+        xstream.allowTypes((String[]) WhitelistObjectInputStream.getWhitelistedClassNames().toArray());
```
decebals: I don't understand why these cast (see diff).

pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java
```diff
@@ -60,7 +64,6 @@ public String toString(Object object) {
         return xstream().toXML(object);
     }
 
-    @SuppressWarnings("unchecked")
```
decebals: Without this annotation I retrieve a warning. The annotation was added by me in the previous commit.

pippo-content-type-parent/pippo-xstream/src/main/java/ro/pippo/xstream/XstreamEngine.java
```diff
@@ -41,17 +45,17 @@ public String getContentType() {
...
         xstream.autodetectAnnotations(true);
-
         // prevent xstream from creating complex XML graphs
         xstream.setMode(XStream.NO_REFERENCES);
 
-        // setup security (see http://x-stream.github.io/security.html)
-        xstream.allowTypes(WhitelistObjectInputStream.getWhiteClassNames());
-        xstream.allowTypesByRegExp(WhitelistObjectInputStream.getWhiteRegEx());
+        // clear out existing permissions and set own ones
+        xstream.addPermission(NoTypePermission.NONE);
```
decebals: So, this is the single modification.

Issue #480, opened 2018-11-26
UndertowWebSocketConnection uses wrong method to close the socket


whyicantusemyemailasusername:
it calls WebSocketChannel.sendClose() which is wrong; javadoc says `Send a Close frame without a payload`. It actually clears up closeReason/closeCode before sending the close frame.
I suppose `WebSockets.sendClose` or `WebSockets.sendCloseBlocking` should be used instead.

decebals:
@whyicantusemyemailasusername 
Can you submit a PR? Thanks!

PR #483, opened 2018-11-29
Issue #480 - Use WebSockets.sendClose to send close code/reason

whyicantusemyemailasusername:
fix for #480


decebals:
Thanks!

decebals:
I didn't know about existence of `WebSockets` class. I see many other useful methods in that class that can be useful to write compact code.

PR #485, opened 2018-11-29
Issue #484 - Use undertow channel tasks to remove closed/broken connections

### LINKED ISSUE #484: UndertowWebSocketAdapter memory leak
UndertowWebSocketAdapter has connections collection keeping list of all undertow connections. However if connection is broken (network error, etc). OnCloseMessage is never called - because there is no close message in ws - and connection is stuck in the list. Also WebSocketContext receives exact reference to connections list. The list can be modified freely and could lead to obscure errors

whyicantusemyemailasusername:
… from connections list.
Undertow-specific Channel.addCloseTask method used to make sure the connection is always removed from the list.

Use readonly wrapper for connections passed to connection context.

fixes #484

### DIFF
--- a/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/websocket/UndertowWebSocketAdapter.java
+++ b/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/websocket/UndertowWebSocketAdapter.java
@@ -33,6 +33,7 @@ import ro.pippo.core.websocket.WebSocketHandler;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
 import java.nio.ByteBuffer;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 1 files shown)


pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/websocket/UndertowWebSocketAdapter.java
```diff
@@ -48,6 +49,7 @@
     private static final Logger log = LoggerFactory.getLogger(UndertowWebSocketAdapter.class);
 
     private static List<WebSocketConnection> connections = new CopyOnWriteArrayList<>();
+    private static List<WebSocketConnection> connectionsReadOnly = Collections.unmodifiableList(connections);
```
decebals: It's for the first time when I see this construction in the context of dynamic `connections` variable. I was wondering if I modified the `connections` list if the modification is propagated to `connectionsReadOnly`. And the response is yes. So, the code works perfect.
For more info see [this](https://stackoverflow.com/questions/15111073/collections-unmodifiablelist-and-defensive-copy) discussion.

decebals:
Excellent. Thanks!

PR #488, opened 2018-12-31
Added a constructor and convenience methods for setting route names d…

Wavesonics:
…uring creation

I was looking at adding the new Admin handlers to my application, and noticed that when adding routes not using the Controller method, there was a bit more friction to naming them.

So here I added a ctor with the route name as an optional parameter. And the various convenience methods for creating routes also have overloads to take in a name.

This way named routes can be created in one step just like unnamed routes.

Note that I didn't add an overload for the ALL method, as naming something created that way seems... wrong?

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/route/Route.java
+++ b/pippo-core/src/main/java/ro/pippo/core/route/Route.java
@@ -38,8 +38,13 @@ public class Route {
     private Map<String, Object> attributes;
 
     public Route(String requestMethod, String uriPattern, RouteHandler routeHandler) {
+        this(requestMethod, null, uriPattern, routeHandler);
+    }
+
+    public Route(String requestMethod, String name, String uriPattern, RouteHandler routeHandler) {
         this.requestMethod = requestMethod;
         this.uriPattern = uriPattern;
+        this.name = name;
         setRouteHandler(routeHandler);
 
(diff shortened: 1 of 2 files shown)


Wavesonics:
It allows named routes as such:
`GET("/admin/ping", "admin.ping", PingHandler())`

decebals:
I don't see a value in this PR.
You can use `named` method like:
```java
GET("/login", routeContext -> routeContext.render("admin/login")).named("adminLogin");
```

See below some examples (including AdminRoutes that is a RouteGroup):
- https://github.com/pippo-java/matilda/blob/master/src/main/java/ro/fortsoft/matilda/MatildaApplication.java#L129
- https://github.com/pippo-java/matilda/blob/master/src/main/java/ro/fortsoft/matilda/web/AdminRoutes.java#L105

I don't prefer telescopic constructors/methods as approach because in this mode you pollute the API (duplicates methods only to add a new parameter).

Wavesonics:
Ha ya, after i saw how much code it added I was questioning it's useful as well. No problem, I think I'll switch my stuff back as well.

decebals:
@Wavesonics 
> Ha ya, after i saw how much code it added I was questioning it's useful as well. No problem, I think I'll switch my stuff back as well.

Thanks for understanding!

Issue #495, opened 2019-02-15
UndertowWebSocketAdapter - race condition in onConnect() call


whyicantusemyemailasusername:
`onConnect()` method calls `WebSocketHandler.onOpen()` to help with websocket initializing. However if initialization process takes some time, `WebSocketHandler` receives `onMessage()` calls (from other threads) _before_ `onOpen()` is finished, leaving handler's state partially initialized. Looks like this is due to `channel.resumeReceives();` call - it should be moved to the end of `onConnect()` method.
One of the workarounds is to put a barrier/semaphore or use CompletableFuture to indicate that initialization process has finished. However it is better to fix this.

PR #496, opened 2019-02-15
Issue #495 - do not call channel.resumeReceives() until handler.onOpen() called

whyicantusemyemailasusername:
Fix for #495 - move channel resume to the end of the `onConnect` method


decebals:
Thanks!

Issue #497, opened 2019-02-15
Java 11 build is broken


whyicantusemyemailasusername:
Due to legacy code removal, some modules are not available anymore.
To reproduce, one can run `mvn clean package` on jdk 11

whyicantusemyemailasusername:
javax.xml.bind.annotation dependencies should be explicitly added to the pom

PR #498, opened 2019-02-16
Issue #497: adopt build for java 11

whyicantusemyemailasusername:
Various fixes to make maven build work on java 11 (see #497) :
- lombok version update. it is used in single file (test); can be removed completely from the project
- add explicit jaxb dependencies where required; see https://stackoverflow.com/questions/43574426/how-to-resolve-java-lang-noclassdeffounderror-javax-xml-bind-jaxbexception-in-j
- fix javadoc issues (closing <p> tags, wrong `@see` syntax, html entities) 
- update mockito versions
- update maven plugin versions

### DIFF
--- a/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerHandler.java
+++ b/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerHandler.java
@@ -462,7 +462,7 @@ public class ControllerHandler implements RouteHandler {
     /*
      * Cleans a complex content-type or accept header value by removing the
      * quality scores.
-     * <p/>
+     * <p>
      * <pre>
      * text/html,application/xhtml+xml,application/xml;q=0.9,image/webp
      * </pre>
--- a/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java
+++ b/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java
@@ -180,10 +180,11 @@ public class ClassUtils {
      * Class} object, including public, protected, default (package)
      * access, and private methods, <b>but excluding inherited methods</b>.
      *
-     * <p>This method differs from {@link Class#getDeclaredMethods()} since it
+     * <p>
+     * This method differs from {@link Class#getDeclaredMethods()} since it
      * does <b>not return bridge methods</b>, in other words, only the methods of
      * the class are returned. <b>If you just want the methods declared in
-     * {@code clazz} use this method!</b></p>
+     * {@code clazz} use this method!</b>
      *
      * @param clazz Class
      *
--- a/pippo-core/src/main/java/ro/pippo/core/Pippo.java
+++ b/pippo-core/src/main/java/ro/pippo/core/Pippo.java
@@ -87,16 +87,12 @@ public class Pippo implements ResourceRouting, ReloadWatcher.Listener {
      * Entry point for a custom WebServer.
      * The idea is to create a custom WebServer if you want to override some aspects (method) of that server or
      * if you want free access to the servlet container (Jetty, Tomcat, ...).
-     *
      * <p>
-     * Show below the code for a <@code>JettyServer</@code> with persistent sessions.
-     * </p>
-     *
+     * Show below the code for a <code>JettyServer</code> with persistent sessions.
      * <pre>
-     * <@code>
      * public class MyJettyServer extends JettyServer {
      *
-     *     @Override
+     *     &#064;Override
      *     protected ServletContextHandler createPippoHandler() {
      *         ServletContextHandler handler = super.createPippoHandler();
      *
@@ -122,7 +118,6 @@ public class Pippo implements ResourceRouting, ReloadWatcher.Listener {
      *     }
      *
      * }
-     * </@code>
      * </pre>
      *
      * @param server
--- a/pippo-content-type-parent/pippo-csv/pom.xml
+++ b/pippo-content-type-parent/pippo-csv/pom.xml
@@ -43,7 +43,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.16.6</version>
+            <version>1.18.6</version>
             <scope>test</scope>
         </dependency>
 
--- a/pippo-controller-parent/pippo-spring/src/main/java/ro/pippo/spring/SpringControllerFactory.java
+++ b/pippo-controller-parent/pippo-spring/src/main/java/ro/pippo/spring/SpringControllerFactory.java
@@ -68,7 +68,7 @@ public class SpringControllerFactory implements ControllerFactory {
     /**
      * Created a bean definition for a class.
      * Optionally configure all bean properties, like scope, prototype/singleton, etc using:
-     * <p/>
+     * <p>
      * <pre>
      * BeanDefinition definition = super.createBeanDefinition(beanClass);
      * definition.setScope(BeanDefinition.SCOPE_SINGLETON);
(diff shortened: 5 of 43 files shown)


pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerHandler.java
```diff
@@ -462,7 +462,7 @@ protected void handleDeclaredThrownException(Exception e, RouteContext routeCont
     /*
      * Cleans a complex content-type or accept header value by removing the
      * quality scores.
-     * <p/>
+     * <p>
```
decebals: Why?

pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java
```diff
@@ -180,10 +180,11 @@
      * Class} object, including public, protected, default (package)
      * access, and private methods, <b>but excluding inherited methods</b>.
      *
-     * <p>This method differs from {@link Class#getDeclaredMethods()} since it
+     * <p>
+     * This method differs from {@link Class#getDeclaredMethods()} since it
      * does <b>not return bridge methods</b>, in other words, only the methods of
      * the class are returned. <b>If you just want the methods declared in
-     * {@code clazz} use this method!</b></p>
+     * {@code clazz} use this method!</b>
```
decebals: Without `</p>`, are you sure?

pippo-core/src/main/java/ro/pippo/core/Pippo.java
```diff
@@ -87,16 +87,12 @@ public WebServer getServer() {
...
      * The idea is to create a custom WebServer if you want to override some aspects (method) of that server or
      * if you want free access to the servlet container (Jetty, Tomcat, ...).
-     *
      * <p>
-     * Show below the code for a <@code>JettyServer</@code> with persistent sessions.
-     * </p>
-     *
+     * Show below the code for a <code>JettyServer</code> with persistent sessions.
      * <pre>
-     * <@code>
```
decebals: https://stackoverflow.com/a/19302439

pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerHandler.java
whyicantusemyemailasusername: otherwise javadoc tool fails. See https://stackoverflow.com/questions/5260368/which-tag-should-be-used-as-paragraph-separator-in-javadoc and https://stackoverflow.com/questions/5260368/which-tag-should-be-used-as-paragraph-separator-in-javadoc

pippo-core/src/main/java/ro/pippo/core/Pippo.java
whyicantusemyemailasusername: this particular entry has incorrect `<@code>` syntax (instead of `${@code}`)

whyicantusemyemailasusername:
> I prefer that this PR to contain only things about Java11. For other things about javadoc please submit a separate PR. Thanks!

I can do that, however w/o javadoc fixes build will fail (looks like javadoc tool from java 11 reports invalid tags like <p/> as errors and aborts the build).

whyicantusemyemailasusername:
actually javadoc checks is feature of java 8, enabled by latest version of maven-javadoc-plugin: https://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete

decebals:
Thanks for clarification! I think that you are right.

whyicantusemyemailasusername:
I've got another branch with non-javadoc changes cherry-picked (and disabled docLink feature) - [here](/whyicantusemyemailasusername/pippo/tree/feature/java_11_build). Do you want me to make a separate PR for it (and decline this one)?

decebals:
In other project initiated by me ([PF4J](https://pf4j.org)), now few days ago someone came with a [PR](https://github.com/pf4j/pf4j/pull/278) about how to make the project build on Java 11. This is a sign that Java 11 is on an ascending trend from adoption point of view. In this context this PR is welcome.

I think that the new dependency introduced by this PR:
```xml
<!-- Jaxb -->
<dependency>
    <groupId>javax.xml.bind</groupId>
    <artifactId>jaxb-api</artifactId>
    <version>${jaxb.version}</version>
</dependency>
```

must be placed in the context of Java 11. So, this dependency will be added only when Java version is 11, using a Maven's profile. As example see [this](https://github.com/dmitry-timofeev/pf4j/blob/193d0693c1a3c8c43a6c9a48dfb293e1b2b925bf/pom.xml#L166).

whyicantusemyemailasusername:
I don't think that moving jaxb dependency into separate profile is a good idea. It will lead to two different build artifacts - one with with jaxb jars (if build on java 11) and another one without.
Former can be run on java 8+, latter can be run on java 8 only - java 9 will issue warnings about missing modules and java 11 will fail as jaxb modules doesn't exist there.
In other words, you'll loose forward compatibility.

whyicantusemyemailasusername:
#499 raised

whyicantusemyemailasusername:
> I don't think that moving jaxb dependency into separate profile is a good idea. It will lead to two different build artefacts - one with jaxb jars (if built on java 11) and another one without.
> Former can be run on java 8+, latter can be run on java 8 only - java 9 will issue warnings about missing modules and java 11 will fail as jaxb modules doesn't exist there.
> In other words, you'll loose forward compatibility.

Example of failure: build server has jdk8, production server has jre 11.

whyicantusemyemailasusername:
closing this in favour of #499

decebals:
>> I don't think that moving jaxb dependency into separate profile is a good idea. It will lead to two different build artefacts - one with jaxb jars (if built on java 11) and another one without.
>> Former can be run on java 8+, latter can be run on java 8 only - java 9 will issue warnings about missing modules and java 11 will fail as jaxb modules doesn't exist there.
>> In other words, you'll loose forward compatibility.

> Example of failure: build server has jdk8, production server has jre 11.

I understand very well what you say but I am not convincied yet that to add a dependency  (for java **8** and java **11**) is the correct solution.
What you say about actual situation when java 8 is used? After the modification proposed by you 
(with new dependency), we have a new jar (from dependency) whith classes that are already present in Java 8 (Java version used in production).
I think that the current version (1.12.0) of Pippo works with Java 11 out of the box, if you add the dependecy to JAXB in your project. JAXB is used in `pippo-core` only in `Error` class and I don't know if it's not a good idea to remove JAXB from `pippo-core` module.

whyicantusemyemailasusername:
You're right. Basically there are several options:
- no jaxb dependency, pippo users are required to add it to their poms manually for java 8.
- jaxb dependency, users are required to exclude the dependency for java 9+.
- drop jaxb dependency.
- two different builds.
- dependency in maven profile.
I fully support 'drop' option. However session-ehcache3 tests fail without jaxb.
Excluding is implemented, for example, in https://github.com/jOOQ/jOOQ/blob/master/pom.xml . For newer VMs there is nothing to do, for older ones - add `<exclude>` section.
No jaxb dependency is the other way around, except users with up-to-date vm would need to perform additional steps.
Dependency in maven profile is most magical one. I'd prefer to fail fast.

decebals:
I think we should go with option `no jaxb dependency, pippo users are required to add it to their poms manually for java 9+` because:
- majority of PF4J users use Java 8 (or don't use Java 9+)
- don't brake actual applications that use PF4J

So, I believe that it's a good idea to add the dependencies in the pom.xml files (current PR) but in a comment form with a specification that they are useful for Java 9+.

Thanks for your patience! This PR is not so simple and we must somehow to take a decision.

whyicantusemyemailasusername:
Sure :) This also means "maven profile for jaxb to able to build on jdk9+". I'll update the other PR

PR #499, opened 2019-02-16
Feature/java 11 build

whyicantusemyemailasusername:
Various fixes to make maven build work on java 11 (see #497, #498 ) :

* lombok version update. it is used in single file (test); can be removed completely from the project
* add explicit jaxb dependencies where required; see https://stackoverflow.com/questions/43574426/how-to-resolve-java-lang-noclassdeffounderror-javax-xml-bind-jaxbexception-in-j
* update mockito versions
* update maven plugin versions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -189,15 +189,6 @@
 
     <profiles>
         <profile>
-            <id>jdk8-build</id>
-            <activation>
-                <jdk>[1.8,)</jdk>
-            </activation>
-            <properties>
-                <additionalparam>-Xdoclint:none</additionalparam>
-            </properties>
-        </profile>
-        <profile>
             <id>travis</id>
             <activation>
                 <property>
--- a/pom.xml
+++ b/pom.xml
@@ -93,6 +93,7 @@
                             <value>bar</value>
                         </property>
                     </javaApiLinks>
+                    <doclint>none</doclint>
                 </configuration>
             </plugin>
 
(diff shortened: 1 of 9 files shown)


whyicantusemyemailasusername:
- added java9+ profiles for jaxb dependencies
- removed java8 profile for doclint as pippo requires java8. also selector `[1.8,)` doesn't work for 9+ jdks.

pom.xml
```diff
@@ -176,15 +188,6 @@
     </modules>
 
     <profiles>
-        <profile>
```
decebals: Do yoy think that we must delete this profile (jdk8) or it's a mistake? The require version for PF4J is java 7.

whyicantusemyemailasusername: Pippo requires 1.8+: `<java.version>1.8</java.version>` The profile was always active (actually should be active but it was not on 9), and the parameter is explicitly set in plugin configuration

decebals: My mistake, sorry. I am working on a feature on PF4J (a project initiated by me that requires Java 1.7) and from this reason I mix things.

decebals:
Thanks!

decebals:
Maybe we must update travis CI config file to add openjdk 11 (see https://github.com/pf4j/pf4j/blob/master/.travis.yml).

whyicantusemyemailasusername:
> Maybe we must update travis CI config file to add openjdk 11 (see https://github.com/pf4j/pf4j/blob/master/.travis.yml).

see #500

PR #510, opened 2019-04-29
chore: upgrade fastjson to latest version

munendrasn:
* http://repo1.maven.org/maven2/com/alibaba/fastjson/1.2.57/
* This fixes fastjson test failures in different timezones
* This also has some performance improvements


decebals:
Thanks!

Issue #407, opened 2017-11-19
Binary render type with byte array


ucin:
It would be good if Pippo can render binary type using `byte[]`. I'm intend to use for rendering Protobuf. Currently is impossible because `ContentTypeEngine.toString(Object object)` only allow String type.

```
public interface ContentTypeEngine {

    void init(Application application);

    String getContentType();

    String toString(Object object);

    <T> T fromString(String content, Class<T> classOfT);

}
```

decebals:
I will investigate your request. If you have some ideas please let me know.

decebals:
I started to work on this feature.
The new signature of `ContentTypeEngine` is:
```java
public interface ContentTypeEngine {

    void init(Application application);

    String getContentType();

    <T> T readObject(Class<T> classOfT, Request request);

    void writeObject(Object object, Response response);

}
```
So, the old `toString` method becomes `writeObject` and the old `fromString` method becomes `readObject`.
The modifications are trivial. For example instead of `send(contentTypeEngine.toString(object))` (https://github.com/decebals/pippo/blob/master/pippo-core/src/main/java/ro/pippo/core/Response.java#L923) we have `contentTypeEngine.writeObject(object, this)`.

You should remark that in readObject and writeObject you have full access to Request and Response (maybe some implementations want to deal with headers).
The actual implementations of `ContentTypeEngine`s are text based. In this case the implementations of readObject and writeObject are trivial, based on `request.getBody()` and `response.send(object.toString())`.

I see a single problem for now. `PippoRule` uses the actual toString and fromString for RestAssured's ObjectMapper (https://github.com/decebals/pippo/blob/master/pippo-test/src/main/java/ro/pippo/test/PippoRule.java#L100) and I don't know for the moment how to resolve this mismatch (I don't have a Request, Response).

Any idea, any advice or feedback is welcome.

lvivtotoro:
Any news on this, so far?  
Just realized Jetty doesn't support partial requests, so I was left with implementing that myself

decebals:
@lvivtotoro 
I didn't finish my work on this feature because I didn't find a solution for the problem with `PippoRule`. I think that I can push my work on GitHub, maybe someone will find a solution.

PR #518, opened 2019-06-05
Add protobuf support

denysvitali:


decebals:
Interesting. This PR is a particular case of #407. I will push my work on GitHub in a separate branch to make my work public.

denysvitali:
Does my work look good? I think this may solve #407 as well, as you said :+1:

denysvitali:
I use it like this, where `LoginResponse` obviously is a Protobuf Message:
```java
ctx.status(HttpStatus.OK_200);
if(req.getUsername().equals("user") && req.getPassword().equals("password")){
resp = LoginResponse.newBuilder()
      .setSuccess(true)
      .setMessage("Login successful")
      .build();
  ctx.protobuf().send(resp);
} else {
  resp = LoginResponse.newBuilder()
      .setSuccess(false)
      .setMessage("Invalid username/password")
      .build();
  ctx.protobuf().send(resp);
}
```

decebals:
> Does my work look good? I think this may solve #407 as well, as you said

I will do an evaluation and I will come with a response. Thanks!

PR #532, opened 2020-02-12
Improved FileItemExtractor

### LINKED ISSUE #531: pippo-controller: there is no way to specify alternative file parameter name
Here the example from pippo.ro site:
```java
@POST("/upload")
@Produces(Produces.TEXT)
public String upload(FileItem file) {..}
```
This code works on java 8+, and only if parameter names are enabled (-parameters argument). There is no way to specify alternative name like one can do for request parameters via @Param("uploaded-file") annotation.

I can think on multiple ways to fix this:
1. Introduce another annotation and use it for such cases (@Multipart("file-name") for example)
1. Introduce method-agnostic way to declare parameter name, like @Named("file-name"), and put name-resolution logic into MethodParameter class.
1. Document such behaviour as "by design"

What do you think?

mhagnumdw:
Fixes #531


mhagnumdw:
Missing to check the code duplication described here https://github.com/pippo-java/pippo/issues/531#issuecomment-584983089-permalink

mhagnumdw:
Hi @decebals !

What do you suggest to avoid duplication described here https://github.com/pippo-java/pippo/issues/531#issuecomment-584983089-permalink ?

😭 Too bad that annotations in Java do not accept interfaces 😃

pippo-content-type-parent/pippo-csv/pom.xml
```diff
@@ -44,7 +44,7 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>1.18.6</version>
-            <scope>test</scope>
+            <scope>provided</scope> <!-- https://projectlombok.org/setup/maven -->
```
mhagnumdw: The scope `provided` was necessary to run the tests and is the recommended scope in the Lombok documentation. See https://projectlombok.org/setup/maven

pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/extractor/FileItemExtractor.java
```diff
@@ -34,9 +37,25 @@ public boolean isApplicable(MethodParameter parameter) {
 
     @Override
     public Object extract(MethodParameter parameter, RouteContext routeContext) {
-        String name = parameter.getParameterName();
+        Param annotation = parameter.getAnnotation(Param.class);
```
whyicantusemyemailasusername: This might not work as expected - see `ControllerHandler.initExtractors()`. annotated FileItem might be processed by both extractors, and the result depends on extractors order.
If you want to reuse @Param annotation, the better way would be to remove FileItemExtractor completely and add additional check `if(FileItem.class.isAssignableFrom(parameterType))` in ParamExtractor.

mhagnumdw: Sorry!
Thanks! Now I see the problem exactly where you said it. I believe that the solution you proposed is simple and makes sense.
I will do some tests in this direction...

mhagnumdw: @whyicantusemyemailasusername , done. Any other suggestions? Please, let me know.

decebals:
@mhagnumdw 
If this PR, in the current form, solves the problem raised by @whyicantusemyemailasusername in #531, you can merge it (with option `Squash and merge` please).
We will see in the near future how we will solve the duplication problem. From my point of view, this PR is OK.

mhagnumdw:
> you can merge it (with option Squash and merge please)

@decebals , it would be a pleasure, but I'm not allowed.

GitHub message:

> Only those with write access to this repository can merge pull requests.

decebals:
@mhagnumdw 
> @decebals , it would be a pleasure, but I'm not allowed.

Done. Please try again.

Issue #530, opened 2020-01-24
Controllers are not being located in subpackages / addControllers


mhagnumdw:
Hi!

### Problem
Controllers are not being located in subpackages in some cases.

### Structure
```
.
└── ro
    └── pippo
        └── classpathtest
            ├── controllers
            │   ├── b
            │   │   └── ControllerB.java
            │   └── ControllerA.java
            ├── PippoApplication.java
            └── PippoLauncher.java
```

### PippoApplication
```java
public class PippoApplication extends ControllerApplication {

    @Override
    protected void onInit() {
        // ...
        addControllers("ro.pippo.classpathtest.controllers");
        // ...
    }

}
```

### Cases
- `ControllerA` is always recognized, we have no problem here
- `ControllerB` is only recognized when the application is run by the JAR like this `java -classpath classpathtest-0.0.1-SNAPSHOT.jar ro.pippo.classpathtest.PippoLauncher`
- `ControllerB` is not recognized when the application is started by an IDE, Eclipse for example, or via the command line with `mvn clean compile exec: java`

### Where's the problem?
Here: [ro.pippo.controller.util.ClassUtils.getClasses(String...)](https://github.com/pippo-java/pippo/blob/release-1.12.0/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java#L70)

In JAR mode [`if`](https://github.com/pippo-java/pippo/blob/release-1.12.0/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java#L77) is executed and in non-JAR mode [`else`](https://github.com/pippo-java/pippo/blob/release-1.12.0/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java#L97) is executed, giving a difference in the result of the method.

The size of this method seems strange to me, but I haven't looked at it yet.

### Sample project
https://github.com/mhagnumdw/classpathtest

### Reproduce
```bash
git clone https://github.com/mhagnumdw/classpathtest.git
cd classpathtest/
```

#### Bad things here (IDE version)
```bash
mvn clean compile exec:java
# or via IDE: Eclipse
```
- It works: http://localhost:8338/a
- **Does not work:** http://localhost:8338/b

#### Everything is beautiful here (JAR version)

```bash
mvn clean package
cd target
[...]

PR #536, opened 2020-04-27
Solves issue #530

mhagnumdw:
Solves #530


decebals:
I like this solution because it does not involve adding a new dependency.

pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java
```diff
@@ -97,33 +98,45 @@
...
+//                    }
                 }
             }
         }
 
+        // TODO: why unmodifiableCollection ?!
         return Collections.unmodifiableCollection(classes);
     }
 
+    private static List<String> getAllFiles(URL packageUrl, String endsWith) {
```
decebals, line 129: If this private method is not used for other purpose (in other places), I propose you to rename it to "getClassFiles" and to remove the `endsWith` parameter.

mhagnumdw:
I'm thinking how to create a test for `ClassUtils.getClasses`.

pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/util/ClassUtils.java
```diff
@@ -97,33 +98,45 @@
...
+                            .collect(Collectors.toList())
+                        );
+//                    } catch (Exception e) {
+//                        throw new PippoRuntimeException(e, "Failed to get classes for package '{}'", packageName);
+//                    }
                 }
             }
         }
 
+        // TODO: why unmodifiableCollection ?!
```
decebals, line 125: I don't remember. From the semantic point of view, I think that is correct: a package contains a static number of classes files.
But sure, you can remove it if you think that a immutable collection is not necessary.

mhagnumdw:
Hi @decebals !

Let me know if you want me to change anything.

I needed to create dummy classes and a ultra small jar inside src/test/resources/ in order to create the test for the `ClassUtils.getClasses` method. Let me know if you want me to change anything, or even take another approach.

decebals:
I am happy to see some tests :smile:. I think that the PR is good. I prefer to not store the classes use din test and the jar file but this is another story. I think that we can improve this aspect in the future.
For example, in [PF4J](https://github.com/pf4j/pf4j) I use with success in tests, the [PluginJar](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/plugin/PluginJar.java). If you have some time and you think that helps, you can copy the `PluginJar` in Pippo, change the name (something like `DynamicJar` or something better) and try to create the jar used in this test, in a dynamic mode, without to have `Class1`-`Class6` and the jar file in repo. You can take a look of [AbstractExtensionFinderTest](https://github.com/pf4j/pf4j/blob/ec03fd022f032abc1669d0ff485c60db5ad1b5c3/pf4j/src/test/java/org/pf4j/AbstractExtensionFinderTest.java) to see how I create class bytes from java sources on the fly via compilation.
Again, if you have time and you like the research. Otherwise the PR is OK for merging.

mhagnumdw:
Hi @decebals !

I agree with you! I'll take a look at what you said.

I saw that your [AbstractExtensionFinderTest](https://github.com/pf4j/pf4j/blob/ec03fd022f032abc1669d0ff485c60db5ad1b5c3/pf4j/src/test/java/org/pf4j/AbstractExtensionFinderTest.java) user `com.google.testing.compile.Compiler.javac()`. I have a minimalist lib that is also used here [TestUtils.java](https://github.com/mhagnumdw/bean-info-generator/blob/a2afe0fe47251364ab0424c6e13f9e2c10d6711f/src/test/java/io/github/mhagnumdw/beaninfogenerator/TestUtils.java#L45).

decebals:
> I have a minimalist lib that is also used here [TestUtils.java](https://github.com/mhagnumdw/bean-info-generator/blob/a2afe0fe47251364ab0424c6e13f9e2c10d6711f/src/test/java/io/github/mhagnumdw/beaninfogenerator/TestUtils.java#L45).

This means that the work is almost ready :smile:. I redirected you to PF4J just to see the code. If you like something from that project, you can use that part here. If you think that it's too complicate for our needs, try to find something more simple.

mhagnumdw:
Hi @decebals !

Today I looked at it again and I have good and bad news.

- Good

Creating the Jar and classes at test run time worked! 🥳 

- Bad

As we are testing a class with static methods I need to use `@RunWith(PowerMockRunner.class)` and this ends up causing the compilation `Compiler.javac().compile(classes)` to break with the message `java.lang.ClassCastException: class com.sun.tools.javac.api.JavacTool`.

If I remove `@RunWith(PowerMockRunner.class)` I don't have the compilation problem, but then I can't mock the static methods I need.

I kind of blindly tried to use the runner `PowerMockJUnit49RunnerDelegateImpl.class`, but then I have the same problem mentioned above.

Another solution, which works, is to hack the `ClassUtils` class, which I basically did:

```java
public class ClassUtils {

    private final static Logger log = LoggerFactory.getLogger(ClassUtils.class);

    private static ClassUtils INSTANCE; // added

    protected ClassUtils() { // added
        ClassUtils.INSTANCE = this;
    }

    @SuppressWarnings("unchecked")
    public static <T> Class<T> getClass(String className) {
        try {
            // return (Class<T>) Class.forName(className); // removed
            return (Class<T>) Class.forName(className, true, INSTANCE.getClassLoader()); // added
        } catch (Exception e) {
            throw new PippoRuntimeException("Failed to get class '{}'", className);
        }
    }

    protected ClassLoader getClassLoader() { // added - It will be mocked in the test
        return this.getClass().getClassLoader();
    }

   // ...
}
```
Within the `ClassUtilsTest` class I create a mock of `ClassUtils`, something like this:

```java
public static class ClassUtilsMock extends ClassUtils {
    private URL jarUrl;
    ClassUtilsMock(URL jarUrl) {
        super();
        this.jarUrl = jarUrl;
    }
    @Override
    protected ClassLoader getClassLoader() {
        return new URLClassLoader(new URL[] { jarUrl });
    }
}
```
And in the test methods I refer to `ClassUtilsMock`.

What do you think? Any comments?

ps: I haven't committed yet.

mhagnumdw:
Ohh, I had forgotten about this:

> https://github.com/powermock/powermock/wiki/FAQ
>
> I get a ClassCastException from DocumentBuilderFactory, SaxParserFactory or other XML related classes
**The reason is that the XML framework tries to instantiate classes using reflection and does this from the thread context classloader (PowerMock's classloader) but then tries to assign the created object to a field not loaded by the same classloader.** When this happens you need to make use of the @PowerMockIgnore annotation to tell PowerMock to defer the loading of a certain package to the system classloader. What you need to ignore is case specific but usually it's the XML framework or some packages that interact with it. E.g. @PowerMockIgnore({"org.xml.*", "javax.xml.*"}). Another option would be to try to bootstrap using our Java Agent.

I spent a lot of time on this!

Adding the annotation below to the test class the problem went away:

```java
@PowerMockIgnore({
    "com.sun.tools.*",
    "javax.tools.*",
})
```

mhagnumdw:
Hi @decebals !

All changes are done. Please let me know if you want me to make any changes.

decebals:
You committed a new `.derived ` file. What is the purpose of this file? I feel that is a misake.

decebals:
The code looks clean. I think that you can merge this request (always with squash :)). I don't know about file `.derived`.

mhagnumdw:
> You committed a new `.derived ` file. What is the purpose of this file? I feel that is a misake.

Hi!

Folders/files that are auto-generated can be marked in Eclipse as `derived`. This causes them to be disregarded by Eclipse, not appearing in the search etc.

This file is from this plugin: https://nodj.github.io/AutoDeriv/

I can remove it without any problems. What do you say?

decebals:
> I can remove it without any problems. What do you say?

Yes. And I think that is a good idea to add a new entry in `.gitignore` to prevent its addition in the future.

PR #542, opened 2020-05-26
Close gzipOutputStream after the usage

### LINKED ISSUE #541: Probable Native memory leak with Java11
Recently, We upgraded one of our service to Java11. We are running the service in Kubernetes. After this upgrade, we started facing restarts because of Kernel OOMKiller.
On debugging, We found Deflator was taking much of the off-heap memory.
All calls to our service have `Accept-Encoding` set to `gzip`. When we looked into GzipResponseStream, we observed GzipOutputStream(this uses Deflater which allocates native memory) is never closed

https://github.com/pippo-java/pippo/blob/a6ab828f9b86c3d6d6737bcb1cd8d1c9d08c9b78/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseStream.java#L45

Not closing GzipOutputStream has lead to native memory in other libraries too - https://issues.apache.org/jira/browse/CURATOR-354

@decebals Is there any reason why `close()` in GzipResponseStream doesn't call `GzipOutputStream`'s close()?

munendrasn:
Fixes #541

### DIFF
--- a/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseStream.java
+++ b/pippo-core/src/main/java/ro/pippo/core/gzip/GZipResponseStream.java
@@ -46,20 +46,20 @@ public class GZipResponseStream extends ServletOutputStream {
         if (closed) {
             throw new IOException("This output stream has already been closed");
         }
+        try (ServletOutputStream outputStream = response.getOutputStream()) {
+            gzipOutputStream.finish();
 
-        gzipOutputStream.finish();
+            byte[] bytes = byteArrayOutputStream.toByteArray();
 
-        byte[] bytes = byteArrayOutputStream.toByteArray();
+            response.addHeader("Content-Length", Integer.toString(bytes.length));
+            response.addHeader("Content-Encoding", "gzip");
 
-        response.addHeader("Content-Length", Integer.toString(bytes.length));
-        response.addHeader("Content-Encoding", "gzip");
-
-        ServletOutputStream outputStream = response.getOutputStream();
-        outputStream.write(bytes);
-        outputStream.flush();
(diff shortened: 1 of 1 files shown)


munendrasn:
Created PR #543 for failing test in master for java11

decebals:
@munendrasn Please test it with your application(s) and if you don't see a downside I am ready to merge it.

munendrasn:
I did some minor testing, things look good. I haven't good perf ests were memory could build up

decebals:
@munendrasn Again, if you think that this PR is ready for merge please merge it. I don't know if we can add a unit test for it. Try it in your application(s) and if the result is OK please merge it.
@mhagnumdw Do you see any problem with the modifications from this PR?

mhagnumdw:
@decebals , looks good to me! I ratify what you said.

💭 ps: I was thinking (just thinking) what a unit test would be like to detect memory leak. From what I understand and to be more precise, I would have to start a new jvm just with this test. 💭

decebals:
@mhagnumdw Thanks for feedback!

> looks good to me! I ratify what you said.

In this case, @munendrasn please press the "Squash and merge" button. It's time for your first PR merge in Pippo :smile:.

>  ps: I was thinking (just thinking) what a unit test would be like to detect memory leak. From what I understand and to be more precise, I would have to start a new jvm just with this test.

I think that it's too complicated so let's do it without a test.

munendrasn:
@decebals Thanks for the merge access

>In this case, @munendrasn please press the "Squash and merge" button

Regarding this, Github allows Repo admin(in this case you) to set branch rules, through which only squash-merge can be allowed, all other options remain disabled

PR #543, opened 2020-05-26
fix failing test in java11

munendrasn:
* ClassUtilsTest fails in java11. This fixes the issue

Fix based on [this](https://stackoverflow.com/questions/50456726/mockclassloader-cannot-access-jdk-internal-reflect-superclass-jdk-internal-refle)

### DIFF
--- a/pippo-controller-parent/pippo-controller/src/test/java/ro/pippo/controller/util/ClassUtilsTest.java
+++ b/pippo-controller-parent/pippo-controller/src/test/java/ro/pippo/controller/util/ClassUtilsTest.java
@@ -15,20 +15,7 @@
  */
 package ro.pippo.controller.util;
 
-import static org.junit.Assert.assertEquals;
-
-import java.io.IOException;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLClassLoader;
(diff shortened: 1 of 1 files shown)


decebals:
It's OK for me.

munendrasn:
@decebals could you please merge this so that I can rebase and rerun PR #542?

decebals:
> @decebals could you please merge this so that I can rebase and rerun PR #542?

Done

decebals:
Thanks for fix!

mhagnumdw:
Oh, sorry! Thanks @munendrasn!!

I was going to take a look at it today, I ended up accepting the MR without seeing it. I only ran local tests with Java 8.

munendrasn:
@mhagnumdw Np :)

Previously, PR used to show the status of Travis builds. So, it was easy to know if build succeed or not. I'm not sure what has changed but now, only coveralls status is shown
If we fix that then, it would become easier to know about build status. Otherwise, we have to directly go to Travis page and check

mhagnumdw:
@decebals, do you know anything about it?

decebals:
I didn't know about this problem. I merged it because the modification looks trivial and I think that the risk to broke something is very low.

PR #544, opened 2020-05-26
Handle non-empty Path and empty method path

### LINKED ISSUE #525: Trailing slashes added from registered routes
Hi!

```java
@Path("/webapp/api/p/releasenotes")
public class ReleaseNotesNgController {
    
    @PUT({ "", "/{id: \\d+}" })
    public void save(@Body ReleaseNotesFormDTO dto) {
        //....
    }

}
```

The routes registered for the above example are:
- Add route for PUT '/webapp/api/p/releasenotes/' **(with trailing slash)**
- Add route for PUT '/webapp/api/p/releasenotes/{id: \d+}'

But I believe they should be:
- Add route for PUT '/webapp/api/p/releasenotes' **(no trailing slash)**
- Add route for PUT '/webapp/api/p/releasenotes/{id: \d+}'

What do you think?

munendrasn:
Fixes #525 

We faced this issue during upgrading from 1.7.0 to the latest Pippo version.

### DIFF
--- a/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java
+++ b/pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java
@@ -34,10 +34,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * Register annotated controller routes.
@@ -67,11 +64,11 @@ public class ControllerRegistry {
      * @param packages
      */
     public void register(Package... packages) {
-        List<String> packageNames = Arrays.stream(packages)
+        String[] packageNames = Arrays.stream(packages)
             .map(Package::getName)
-            .collect(Collectors.toList());
+            .toArray(String[]::new);
 
-        register(packageNames.toArray(new String[packageNames.size()]));
+        register(packageNames);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java
```diff
@@ -195,11 +192,31 @@ private void registerControllerMethods(Map<Method, Class<? extends Annotation>>
                 // add an empty string to allow controllerPaths iteration
                 controllerPaths.add("");
             }
+            if (methodPaths.length == 0) {
```
decebals, line 195: I see the same construction for `controllerPath`.  Is it not possible to remove these blocks and to make the code more readable?   For example `if (controllerPaths is empty) { // no controlers => return }`. The same for `methodPaths`.

decebals:
@munendrasn I am happy to see that you added some unit tests :+1:.

pippo-controller-parent/pippo-controller/src/main/java/ro/pippo/controller/ControllerRegistry.java
munendrasn: I could make this change but complexity would remain same. As user could define `@Path("")` or `@GET("")` in which case, controllerPath or methodPath would be non-empty.

I could covert the above using ternary operator
```java
controllerPaths = controllerPaths.isEmpty? Collections.singletonSet(""): controllerPaths
methodPaths = methodPaths.length == 0? new String[]{""}: methodPaths
```
does this look good or I misunderstood?

decebals:
@munendrasn I think that it's ready for merge if @mhagnumdw doesn't see other problems.
@munendrasn Because now you are a committer (congratulations :1st_place_medal:) please merge it yourself (always use `Squash and merge` option because we are not interested in keeping the intermediate committees)

munendrasn:
I will wait for @mhagnumdw's feedback and then merge it

mhagnumdw:
@munendrasn , go ahead.

PR #555, opened 2020-07-02
Add possibility to disable suffix for Pebble loader

decebals:
### Problem

I cannot use the excellent [Pebble IntelliJ plugin](https://github.com/bjansen/pebble-intellij) to navigate on `include/extends`:

```
{% extends "base" %}
```

If I disabled the suffix (`.peb` that is added to the template name) I can navigate with CTRL + MOUSE CLICK to referenced template
 
```
{% extends "base.peb" %}
```

### Solution

The easy mode is to add a new `pebble.suffix.enabled` in `application.properties` as:
```
pebble.suffix.enabled = false
```

Of course is you disable the suffix (`.peb`) you render the template with:

```java
 Map<String, Object> model = new HashMap<>();
 model.put("users", ticketingService.getUsers());
 routeContext.render("users.peb", model);
```

instead 

```java
 Map<String, Object> model = new HashMap<>();
 model.put("users", ticketingService.getUsers());
 routeContext.render("users", model); // <<< WITHOUT `.peb` file extension
```


decebals:
The solution is trivial and the disable flag is false by default.

PR #562, opened 2020-07-27
Test improvement

eas5:
This is a test refactoring.

**Problem:**
Conditions within the test method will alter the behavior of the test and its expected output, and would lead to situations where the test fails to detect defects in the production method since test statements were not executed as a condition was not met.

**Solution:**
Instead of using a conditional to control the test execution, we use the proper JUnit API with assertNotNull().

**Result:**
_Before:_
if (url != null) {
   assertTrue("Path traversal security issue", Paths.get(url.toURI()).startsWith(basePath));
}
_After:_
Assert.assertNotNull(url);
assertTrue("Path traversal security issue", Paths.get(url.toURI()).startsWith(basePath));

### DIFF
--- a/pippo-core/src/test/java/ro/pippo/core/route/PathTraversalResourceHandlerTest.java
+++ b/pippo-core/src/test/java/ro/pippo/core/route/PathTraversalResourceHandlerTest.java
@@ -15,7 +15,6 @@
  */
 package ro.pippo.core.route;
 
-import org.junit.Assert;
 import org.junit.Test;
 
 import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


decebals:
Thanks!

munendrasn:
After this commit Travis build is failing. Looking at the test-case, the `url` will always be null. So, we should assertNull and remove others
@decebals 
Is there a reason for adding `url!=null`?

decebals:
@munendrasn 

I don't remember about this test in particular. If you can fix the Travis fail please do it. Thanks!

PR #592, opened 2021-11-30
Remove controller listeners

decebals:
I think that they are not fully implemented and used.


decebals:
The first two commits are not related to current subject but I think that we can ignore them (they are trivial changes).

PR #612, opened 2022-11-21
Bumped pac4j version to 5.7.0 and mockito version to 4.9.0

sailingbonbini:
Bumped pac4j version to 5.7.0. The pac4j interfaces changed - dropped the use of generics and added support for java.util.Optional.

### DIFF
--- a/pippo-security-parent/pippo-pac4j/pom.xml
+++ b/pippo-security-parent/pippo-pac4j/pom.xml
@@ -14,7 +14,8 @@
     <description>Pac4j integration</description>
 
     <properties>
-        <pac4j.version>2.2.0</pac4j.version>
+        <pac4j.version>5.7.0</pac4j.version>
+        <mockito.version>4.9.0</mockito.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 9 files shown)


sailingbonbini:
I'm working on a new, green field project and would like to use Java 11.
Happy to depend on 2.x-SNAPSHOT of pippo. Do you have a timeframe for 2.x
in mind?

sailingbonbini:
Alternatively, I could create another branch and bump it to 4.x/Java 8 ?

decebals:
> I'm working on a new, green field project and would like to use Java 11. Happy to depend on 2.x-SNAPSHOT of pippo. Do you have a timeframe for 2.x in mind?

I think it will be early 2023, but I don't have a clear date in mind.
[This](https://github.com/pippo-java/pippo/milestone/1) is what we have in mind for pippo 2.x

> Alternatively, I could create another branch and bump it to 4.x/Java 8 ?

I prefer this variant if it's not too complicated for you. On this occasion we are also doing an incremental transition and it will be easy for us to migrate from 4.x to 5.x

sailingbonbini:
I've added a new PR to bump pac4j to 4.5.

Is there a branch where the work for pippo-2.x is happening? I could create a PR to upgrade pac4j to 5.7. Also, would love to contribute to piccolo-2.x

PR #634, opened 2023-02-15
Dynamically determine free port for Redis to listen on

### LINKED ISSUE #632: JedisSessionDataStorageTest fails on machine that has Redis running
Caused by `JedisSessionDataStorageTest.setUpClass` using default constructor of `RedisServer` which tries to bind to default port `6379`. If that port is taken (by another Redis running), this "test server" can not be started and test fails:

https://github.com/pippo-java/pippo/blob/master/pippo-session-parent/pippo-session-jedis/src/test/java/ro/pippo/session/jedis/JedisSessionDataStorageTest.java#L39

Instead, test case should find an free port and initialize `RedisServer` using that port.

I'll try to submit a PR.

gkresic:
Prevents tests from failing if default port 6379 is alredy in use.

Closes #632

### DIFF
--- a/pippo-session-parent/pippo-session-jedis/src/test/java/ro/pippo/session/jedis/JedisSessionDataStorageTest.java
+++ b/pippo-session-parent/pippo-session-jedis/src/test/java/ro/pippo/session/jedis/JedisSessionDataStorageTest.java
@@ -16,10 +16,13 @@
 package ro.pippo.session.jedis;
 
 import java.io.IOException;
+import java.net.ServerSocket;
+
 import org.junit.AfterClass;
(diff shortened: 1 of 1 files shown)


gkresic:
@decebals I had one commit **after** you merged this PR, could you see if you can merge that last commit, too? Stupid bug, I forgot to initialize `JedisPool` to the same port as `RedisServer`. Tests on my machine were passing since I had Redis running externally, but failed when run on GutHub Actions (as they should).

decebals:
@gkresic It was my mistake after all, I was in a hurry. Please create a new PR. When the PR is ready for merge please add me as reviewer or give me a sign (a comment).

